PR #182, opened 2019-03-13
Generalise write geotif

sixy6e:
Nothing concrete at this stage.  Just some ideas and future concepts that could be handy down the line.

The Conversion class is an idea to enable a "driver" like concept for handling the I/O of different formats.
eg

* GAMMA raw binary
* GAMMA TIFF
* ROIPAC raw binary
* SNAP BEAM-DIMAP
* SNAP HDF5

The other functions are simplifying already existing functions.  Just written in a way that they're more generic, and only do one base thing, like write a file.  Not prep metadata, or figure out what kind of file i am, which occurs all during the write phase.  So just some ideas on how to split those components out.

I've given examples for both rasterio and gdal-python bindings, and should work under Python2.7 and Python3.  No tests, but the goal was more to present some ideas and concepts, so I wouldn't expect travis to pass as not all imports are available (and some could be wrong).

### DIFF
--- a/pyrate/conversion/base.py
+++ b/pyrate/conversion/base.py
@@ -148,6 +148,12 @@ def write_geotiff(data, out_fname, dtype='float32', metadata=None, crs=None,
 class Conversion(object):
 
     """
+    *****************************************************************
+    NOTE!
+    This presents a concept for a future direction when there is a
+    greater need to support more formats.
+    *****************************************************************
+
     The base conversion class.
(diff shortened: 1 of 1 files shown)


mcgarth:
Thanks for the contribution @sixy6e! I like the idea of using a conversion class to generalise the import of different input data. Perhaps we can revisit this suggestion during the next period of PyRate development

PR #186, opened 2019-07-10
Limit travis builds

brenmous:
This will only trigger travis to run on develop, master and release- (assuming we'll use the "release" prefix for dev -> master staging branches). This can be changed in .travis.yml.


brenmous:
It'll still run on pull requests, which is handy.

PR #189, opened 2019-07-11
Remove all references to Luigi

sheecegardezi:
Remove all Luigi imports
Remove all methods using Luigi imports 
Remove Luigi from installation processes 
Remove all tests for  Luigi methods


mcgarth:
I agree with @brenmous: lets get all the code that used Luigi removed

PR #194, opened 2019-07-18
Package deprecation.

brenmous:
Remove unused python, apt and Raijin package/modules.


brenmous:
Need to test something.

PR #196, opened 2019-07-19
Cleaning top level directory, general housekeeping.

brenmous:
I've moved some things around, if you have any thoughts about how to do it differently feel free to make changes. The one thing I'm set on is getting rid of setup.cfg because it's totally superfluous now. Otherwise it's mostly minor things like moving the contents of HPC to docs and utils and putting the default config files in the top level.

I've also added our changes so far for the 0.3.0 release to HISTORY.rst. I like keeping a detailed changelog as it's easier to match changes to tagged releases compared to trawling git logs, and otherwise it's just a file sitting there doing nothing. I roughly follow the style of https://keepachangelog.com/en/1.0.0/. Let me know what you think.

Also I branched this off my bren-conda branch, it shouldn't matter but review that PR first.


brenmous:
I'll close this till the other one is merged.

PR #207, opened 2019-07-26
Release 0.3.0

brenmous:
- Deprecate Python 2 support, currently only Python 3.6 is now supported
- Restructure documentation (remove installation from README.md, move to /docs)
- Update installation instructions for Ubuntu and NCI
- Update all Python, OS and NCI dependency versions, pin versions
    - This fixes the errors caused by version >2 of Networkx
- Add Docker support
- General housekeeping and cleaning up of project structure
- Deprecate use of Luigi package


mcgarth:
@sheecegardezi there are still lots of references to luigi in test code. I found this by doing:

grep luigi tests*
grep LUIGI tests*

please fully deprecate these tests involving Luigi now that it is removed from package dependencies.

PR #208, opened 2019-07-29
Remove ref to Luigi 

sheecegardezi:
Remove tests and contents referring  to Luigi


sheecegardezi:
changing branch to merge too.

PR #215, opened 2019-08-05
Implement converttogeotiff and prepifg restructure.

brenmous:
This implements Matt's changes, which separates geotiff conversion from prepifg. This adds a new a step to the workflow, `converttogeotiff`, which will convert the IFG unws to tif and place them in the same directory (OBS_DIR in the parameters).

My changes involved tweaking tests to accommodate this (converting tifs before running prepifg) and adding a test module for the conversions step. 

It's important to note that tifs won't be recomputed. You have to remove the converted tifs from the observations directory if you want them to be reconverted.

Documentation still needs to be updated but I wanted to merge this into the codebase so you can start making changes with the new workflow implemented.


brenmous:
Working on the Travis build errors.

brenmous:
Travis is failing because a new numpy version came out and Travis installs it by default. For now we can fix it by force installing numpy 1.16.4 before running the tests - uninstalling as part of the setup didn't seem to work. We'll need to work out a better solution.

PR #223, opened 2019-08-25
Coherence masking

brenmous:
Coherence masking has been implemented. This affect `converttogeotiff` and `prepifg`. 

If masking is enabled, `converttogeotiff` will convert the coherence images to .tif files. During `prepifg`, the interferogram will be maksed before resampling or cropping is performed. This masking occurs on the in-memory copy of the interferogram, so the ifg provided by the user isn't modified.


input_parameters.conf
```diff
@@ -41,6 +41,15 @@ nan_conversion: 1
...
 processes: 8
 
+#------------------------------------
+# Coherence masking options: used by process
+# cohmask: 1 = ON, 0 = OFF
+# cohthresh: coherence threshold value
+# cohfiledir: directory containing the coherence files
+cohmask:   0
+cohthresh:  0.6
+cohfiledir:
```
sheecegardezi: Can you rename constants "cohfiledir" in input_parameters.conf and	"COH_DIR" defined in configh.py to be same. It helps tracing back to the value.  And add a comment by default cohfiledir has a None value or set a default value.

pyrate/gdal_python.py
```diff
@@ -274,8 +300,20 @@ def crop_resample_average(
     tmp_ds = gdal.GetDriverByName('MEM').CreateCopy('', dst_ds) \
         if (match_pyrate and new_res[0]) else None
 
+    src_ds, src_ds_mem = _setup_source(input_tif)   
+
+    if coh_path and coh_thresh:
+        coh_raster = prepifg.dem_or_ifg(coh_path)
```
sheecegardezi: Its a general observation; when I started reading the code "coh" confused me, now that i know its abbreavation of Coherence its a breeze. I think it will be a good idea to expand it coh to coherence in var names.

pyrate/gdal_python.py
```diff
@@ -17,18 +17,44 @@
...
 gdal.SetCacheMax(2**15)
 GDAL_WARP_MEMORY_LIMIT = 2**10
 LOW_FLOAT32 = np.finfo(np.float32).min*1e-10
 
-
+def coherence_masking(src_ds, coh_ds, coh_thresh):
+    """
+    Perform coherence masking on raster in-place. 
+  
+    Based on gdal_calc formula provided by Nahidul:
```
sheecegardezi: Can you add Nahidul to docs/authors.rst as well.

pyrate/config.py
```diff
@@ -466,12 +468,62 @@ def original_ifg_paths(ifglist_path):
...
+    if len(coh_path) == 0:
+        raise IOError(f"No coherence files found for ifg with epoch " 
+                      f"{epoch}. Check that the correct coherence files "
+                      f"exist in {coh_dir}.")
+    elif len(coh_path) > 1:
+        raise IOError(f"Found more than one coherence file for ifg with epoch "
+                      f"{epoch}. Check that the correct coherence files "
+                      f"exist in {coh_dir}. Found:\n {coh_path}")
+    else:
+        return coh_path[0]
```
sheecegardezi: Can you add an example of expected coherence file name in comments.

input_parameters.conf
brenmous: I've changed the lookup key to "COH_FILE_DIR"

pyrate/gdal_python.py
brenmous: Fair point, I've expanded 'coh_' to 'coherence_'

pyrate/gdal_python.py
brenmous: Done!

pyrate/config.py
brenmous: And done

brenmous:
Let's hold off on merging this one till the INSAR team have had a chance to test it and are happy with the output.

input_parameters.conf
brenmous: and added a note about OBS_DIR being the default value

PR #235, opened 2019-09-18
Implement file lists for coherence and header files.

brenmous:
Rather than relying on extensions to identify file types we will now use lists for coherence and GAMMA header files in the same way as intergerogram file lists.

The new parameters are `slcfilelist` and `cohfilelist.`

Another addition is that `cohfiledir` and `slcFileDir` will fallback on `obsdir` as a default if they aren't provided.

Usage documentation has been updated to reflect this.

### DIFF
--- a/pyrate/core/config.py
+++ b/pyrate/core/config.py
@@ -509,9 +509,12 @@ def coherence_path_for(path, params, tif=False) -> str:
                       f"exist in {coherence_dir}. Found:\n {coherence_name}")
     else:
         if tif:
-            coherence_name = coherence_name[0] + '.tif'
+            name, ext = os.path.splitext(coherence_name[0])
+            tif_ext = ext.replace('.', '_') + '.tif'
+            coherence_name = name + tif_ext
         else:
             coherence_name = coherence_name[0]
+
         return os.path.join(coherence_dir, coherence_name)
 
 def coherence_paths(params) -> List[str]:
--- a/pyrate/core/config.py
+++ b/pyrate/core/config.py
@@ -479,11 +479,11 @@ def original_ifg_paths(ifglist_path):
 def coherence_path_for(path, params, tif=False) -> str:
     """
     Returns path to coherence file for given interferogram. Pattern matches
-    based on an expected filename of {epoch}*{extension}.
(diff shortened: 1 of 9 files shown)


pyrate/core/config.py
```diff
@@ -525,11 +527,6 @@ def coherence_paths(params) -> List[str]:
         A list of full paths to coherence files.
     """
     ifg_file_list = params.get(IFG_FILE_LIST)
-    if ifg_file_list is None:
```
sheecegardezi: This check has moved to config?

pyrate/core/gamma.py
```diff
@@ -17,16 +17,15 @@
...
 import re
 import os
-import glob2
 from datetime import date, time, timedelta
 import numpy as np
 import pyrate.core.ifgconstants as ifc
 from pyrate.core import config as cf
 
+
 PTN = re.compile(r'\d{8}')  # match 8 digits for the dates
```
sheecegardezi: I am concerned if more regex creeps in we would lose track of the reason for using these patterns. Can we move this to either constant file and import it here or expand the name to something on the line of "regex pattren 8 digits" and update comments to 8 digits represent the epoch time in the file name.

pyrate/core/gamma.py
```diff
@@ -230,24 +221,34 @@ def get_header_paths(input_file, slc_dir):
...
     """
     _, file_name = split(input_file)
-    matches = PTN.findall(file_name)
-    headers = [glob2.glob(join(slc_dir, '**/*%s*slc.par' % m))[0]
-            	for m in matches]
-    return headers
-
+    epochs = PTN.findall(file_name)
+    header_names = cf.parse_namelist(slc_file_list)
+    matches = [hdr for hdr in header_names if any(e in hdr for e in epochs)]
```
sheecegardezi: An error is thrown at some point if an epoch is not found in the header list right?
```
# along the lines of :
if not any(e in epochs for e in hdr ):
    Exception("header not available for an epoch")
```

brenmous:
Thanks for picking up on those, will implement these changes as part of the bren-config PR (config file validation).

PR #236, opened 2019-09-23
Input parameter validation

brenmous:
This implements a large amount of validation for input parameters.

The majority of the code is in `config.py` - it's now at 1352 lines so we should refactor this stuff out into a validation specific module. I attempted this but to avoid circular imports it would require the configuration constants to be specified in a separate constants module. They are peppered all throughout the codebase so I ditched that idea for the time being.

I'm hoping the validation code is self-explanatory. When the config gets parsed there's now an additional `validate_parameters` function that gets called. This calls validation subroutines (I've tried to group validations together logically). 

Basic bounds checking/file existence etc. is performed by dictionaries of lambda functions - there is one for 'compulsory' parameters and a separate dictionary for each lot of optional parameters. The optional validators are only called if that feature is enabled.

Some more complex validation that relies on interrelated parameters is also called. Note that some validation requires interferograms to exist as geotiffs so we can extract metadata - this is okay as this validation isn't required by converttogeotiff which is the only step where geotiffs won't exist.

This is a big change so any questions feel free to ask.


brenmous:
Working on validating reference pixel search windows

brenmous:
@sheecegardezi I haven't removed the runtime error checking as we discussed but in future if you see error checking throughout the code feel free to check if it's covered by the parameter validation and remove if. My opinion is that the runtime error checking should be considered separate from the config validation so that functionality can be used in isolation but this is your call.

PR #237, opened 2019-09-25
Refactor reference phase estimation code out of process.py and into ref_phs_est.py

### LINKED ISSUE #170: Consolidate Ref_phs_est code
A lot of reference phase estimation code from the ref_phs_est.py module is duplicated in the run_pyrate.py script. As a result the code in ref_phs_est.py is not used during the workflow and is basically useless.

- tests run on code in ref_phs_est.py
- code in run_pyrate.py is optimised for MPI
- code for the most part is nearly identical in the two locations

It would be ideal to build MPI functionality into ref_phs_est.py functions and deprecate those in run_pyrate.py

brenmous:
Closes #170 

I've moved everything related to reference phase estimation, including top level wrapper function and saving np array to disk, to `ref_phs_est.py.` The code in that module retains parallelism via joblib but methods 1 and 2 are now MPI operable. `ref_phs_est.estimate_ref_phs` is the main function. It can take a list of Ifg objects or paths.

To make my life easier, I changed `shared.check_correction_status` to take a a list of IFG objects or paths, rather than the `preread_ifg` dictionary it was taking before.


mcgarth:
There's been a change to screen messages for orbital correction.

What used to say:

+9s pyrate.core.orbital:INFO Removing orbital error using NETWORK correction method and degree=QUADRATIC

Now says:

+9s pyrate.core.orbital:INFO Removing orbital error using 2 correction method and degree=2

Can we fix this back to the verbose (first) way of reporting to screen please @brenmous ?

PR #240, opened 2019-10-04
Validation refactor

brenmous:
This aims to solve the issue of needing access to cropped/resampled interferograms for certain parameter checks. It does this by introducing the previously discussed idea of selectively running validation functions based on the current workflow step being run. _Note that bounds checking is still performed for every parameter regardless of what step is being run_ - this is an improvement that can be made in future.

The main thing to test here is that the previous failing checks (orbital looks, refx/y) now work.

### DIFF
--- a/input_parameters.conf
+++ b/input_parameters.conf
@@ -74,13 +74,12 @@ ifgylast:     -34.22
 
 #------------------------------------
 # Reference pixel search options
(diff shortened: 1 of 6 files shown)


brenmous:
I need to fill out/update some of the docstrings, so hold off on merging till that's been pushed.

PR #245, opened 2019-10-28
 bugfix: Prepifg

sheecegardezi:
- only _check_crop_coords if CUSTOM_CROP are supplied


mcgarth:
Seems to be working as expected. Merging with release-0.4.0

PR #251, opened 2020-02-02
Update Config module

sheecegardezi:


sheecegardezi:
Need to update all the vars accessing paths using var from config module directly first

PR #256, opened 2020-03-30
Travis CI 

basaks:


basaks:
Working travis without the docs generation and deploy stage. Will work on that next.

PR #260, opened 2020-05-06
Mg/port from test branch

mcgarth:
This PR brings in some of the minor differences from the 'test' branch.

Some of the more complicated differences are discussed separately


mcgarth:
I compared the Travis test failures in this PR to the sb/restart_from_master. They seem to be identical, so the issue is not in this PR ;-)

PR #262, opened 2020-05-14
Sb/prepifg investigation

basaks:


mcgarth:
In this PR:
-  the method of using GDAL system calls in prepifg (activated by ``largetifs: 1``) is improved, but not totally fixed. There are still some minor numerical differences between this method and the direct python prepifg method. As a result, ``largetifs: 1`` will be removed from user visibility in the release.
- the number of regression tests is reduced, in order that Travis tests pass within the 50 minute runtime limit.

PR #264, opened 2020-05-16
Refactor user experience

mcgarth:
- remove the user_experience module - move functions to other modules
- rename function for factorising integer
- add tests for factorising integer function
- stop deleting tsincr files


mcgarth:
@basaks please check that the entries in doc/history.rst are accurate for this coming release

PR #266, opened 2020-05-18
added to the change list

basaks:


mcgarth:
I made some minor edits and force-pushed. Also added today's date as release date...!

PR #267, opened 2020-05-23
Sb/ro prepifg

basaks:


.travis.yml
```diff
@@ -78,7 +78,7 @@ deploy:
   verbose: true
   on:
     branch: master
-    condition: $GDALVERSION="3.0.2" && $TRAVIS_PYTHON_VERSION=3.8.*
```
mcgarth, line 82: Does the deleted line not ensure that only one of the 4 travis runs builds the deployment? Don't we need that?

basaks: There is only one python 3.8 Travis env meaning only one build will push the docs in master.

mcgarth:
I have tested this PR against the small test data, and some bigger Sentinel-1 data. The latter of which I had previously encountered a read-only error message for input tif files when prepifg tries to write to its header. I no longer get this behaviour indicating that prepifg is not trying to write to its input data any longer ;-)

PR #268, opened 2020-05-27
Sb/ref phs investigation

basaks:


mcgarth:
I have checked this PR against the bug Iwas encountering with refest=1 using the smal_test data when cohmask=1 and cohtresh =0.5. The original crash I was encountering no longer occurs.

I would like @tfuhrmann to test this PR against the issues he reported

pyrate/core/ref_phs_est.py
```diff
@@ -61,18 +62,15 @@ def _inner(ifg_paths):
...
             for n, ifg in enumerate(ifgs):
                 ifg.phase_data -= ref_phs[n]
         else:
             ref_phs = np.zeros(len(ifgs))
             for n, ifg in enumerate(ifgs):
-                ref_phs[n] = \
-                    _est_ref_phs_method2(phase_data[n], half_chip_size,
-                                         refpx, refpy, thresh)
+                ref_phs[n] = _est_ref_phs_method2(phase_data[n], half_chip_size, refpx, refpy, thresh)
                 ifg.phase_data -= ref_phs[n]
```
mcgarth, line 74: I wonder if we should pull the actual ref_phs subtraction step (and metadata update) from this function, which is 'estimating' the ref phase? it could be a small separate function here or in process.py.
So this function, ref_phs_est_method2 (and also ref_phs_est_method1) just return the ref_phs values. They do not apply the ref_phs correction (subtraction) That is applied elsewhere (i.e. in the mainline, process.py)

tfuhrmann: I'll do some testing today.
Your suggestion @mcgarth  is very much in line what Chandra and I have been proposing anyway for all the corrections (reference phase, orbit, atmosphere): 
1. calculate the correction values and save them in separate files (e.g. npy arrays)
2. subtract the correction from the interferometric phase elsewhere
This allows the user to re-run the calculation of corrections with different parameter settings without having to go back to the 'prepifg' step.

mcgarth: Good point @tfuhrmann ! ref_phs is particularly easy in this regard, since it's just one number per interferogram!

PR #272, opened 2020-06-15
Sb/refpixel lat lon

basaks:
Adds lat/lon refpixel and tests around it


tfuhrmann:
I've re-run my real data test using the latest commit (rounding). The lat/lon to pixel coordinate conversion is now delivering exactly the same output as gdallocationinfo for the first reference area, but still off by one pixel for the second reference area (see below command line output). I think the small difference of one pixel for some locations is accectable.

PyRate command line output: 
ref area 1: Converted reference pixel coordinate (x, y): (563, 227)
ref area 2: Converted reference pixel coordinate (x, y): (117, 501)

PR #276, opened 2020-06-27
Optional save of tsincr files

mcgarth:
- make saving tsincr files (npy or tif) optional.
- controlled by new config parameter `savetsincr`
- default is on (`savetsincr=1`)
- saving of tscuml files is unaffected.


tfuhrmann:
I've run two PyRate runs using "savetsincr: 0" and "savetsincr: 1" on a test data set and can confirm that the results of both runs finished without errors and generated the results as expected.

PR #282, opened 2020-07-26
Sb/step2 refactor process

basaks:
1. remove unittest suite, and use pytest exclusively
2. _prepifg_ outputs readonly multilooked files.
3. _process_ treats multilooked files as readonly
4. write refpixels to disc and reuse


basaks:
See stack outputs are the same before and after this change:

image

See timeseries outputs are the same before and after this change:

image

These are generated using in both _develop_ and the current branch: 
>> pyrate workflow -f tests/test_data/small_test/conf/pyrate_gamma_test.conf

PR #283, opened 2020-07-28
Release 0.4.3

mcgarth:
Added
+++++
- Ability to define the order of steps in the ``process`` workflow
  (default order unchanged).

Fixed
+++++
- Nil

Changed
+++++++
- ``prepifg`` output interferograms are saved as read-only files.
- ``process`` makes a writable copy of the ``prepifg`` output data
  at the beginning of each run.
- The selected reference pixel is saved to disk and re-used on subsequent
  ``process`` runs.
- Saving of incremental time series (``tsincr``) products is optional,
  controlled by the ``savetsincr`` configuration parameter (default is on).

Removed
+++++++
- Removed obsolete InSAR terminology from code, docs and test data
  (changed to `first` and `second` images).
- Stopped using ``unittest`` unit test framework in favour of exclusively
  using ``pytest``.


mcgarth:
Thanks for your reviews @adeane-ga @chandra2ga @tfuhrmann

PR #284, opened 2020-08-01
Sb/step3 refactor process 2

basaks:
1. Extensive testing of refpixel write on disc and reuse
2. Extensive testing of orbital error write on disc and reuse
3. Decluttered orbital.py module.
4. Updates to both refpixel and orbital error files names on disc to include all relevelant params to reduce user oversight requirements

### DIFF
--- a/input_parameters.conf
+++ b/input_parameters.conf
@@ -111,10 +111,12 @@ refest:        2
 # orbfitmethod = 1: interferograms corrected independently; 2: network method
 # orbfitdegrees: Degree of polynomial surface to fit (1 = planar; 2 = quadratic; 3 = part-cubic)
 # orbfitlksx/y: additional multi-look factor for orbital correction
+orbfit:        1
 orbfitmethod:  2
 orbfitdegrees: 1
 orbfitlksx:    1
 orbfitlksy:    1
+orbfitoffset:  1
 
 #------------------------------------
 # APS spatial low-pass filter parameters
--- a/input_parameters.conf
+++ b/input_parameters.conf
@@ -111,7 +111,6 @@ refest:        2
 # orbfitmethod = 1: interferograms corrected independently; 2: network method
 # orbfitdegrees: Degree of polynomial surface to fit (1 = planar; 2 = quadratic; 3 = part-cubic)
 # orbfitlksx/y: additional multi-look factor for orbital correction
-orbfit:        1
 orbfitmethod:  2
 orbfitdegrees: 1
 orbfitlksx:    1
--- a/pyrate/configuration.py
+++ b/pyrate/configuration.py
@@ -102,13 +102,18 @@ class MultiplePaths:
                 converted_path.stem + '_' + str(ifglksx) + "rlks_" + str(ifgcropopt) + "cr.tif")
         self.tmp_sampled_path = tempdir.joinpath(self.sampled_path.name).as_posix()
         self.converted_path = converted_path.as_posix()
-        self.orbfit_correction_on_disc = Path(params[cf.OUT_DIR], cf.ORB_ERROR_DIR,
-                                              self.sampled_path.stem + '_' +
-                                              '_'.join([cf.ORBITAL_FIT_METHOD,
-                                                        cf.ORBITAL_FIT_DEGREE,
-                                                        cf.ORBITAL_FIT_LOOKS_X]) +
-                                              '_orbfit.npy')
         self.sampled_path = self.sampled_path.as_posix()
+
+    @staticmethod
+    def orb_error_path(ifg_path: Union[str, Path], params) -> Path:
+        if isinstance(ifg_path, str):
+            ifg_path = Path(ifg_path)
+        return Path(params[cf.OUT_DIR], cf.ORB_ERROR_DIR,
+                    ifg_path.stem + '_' +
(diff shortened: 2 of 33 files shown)


basaks:
Evidence of unchanged output in develop using the `pyrate_gamma_test.conf `.

Commands used:

```
 1020  git checkout sb/step3-refactor-process-2 
 1021  git pull
 1022  rm -rf out
 1023  mpirun -n 4 pyrate workflow -f tests/test_data/small_test/conf/pyrate_gamma_test.conf 
 1026  git branch
 1027  mv out/gamma/ out/gamma_new
 1028  git checkout develop
 1029  git pull
 1031  mpirun -n 4 pyrate workflow -f tests/test_data/small_test/conf/pyrate_gamma_test.conf 
 1033  for file in `ls out/gamma_new/out/ts*.tif`; do echo $file; gdalinfo $file -stats | grep STATISTICS; done > branch.txt
 1034  for file in `ls out/gamma/out/ts*.tif`; do echo $file; gdalinfo $file -stats | grep STATISTICS; done > develop.txt
```

Attached outputs.
[branch.txt](https://github.com/GeoscienceAustralia/PyRate/files/5013729/branch.txt)
[develop.txt](https://github.com/GeoscienceAustralia/PyRate/files/5013730/develop.txt)

mcgarth:
> Why does this work in the test configs?

because the test configs (that I can see) only have one instance of `orbfit:` assignment. I think you didn't notice that `orbfit:` was moved to the top of `input_parameters.conf` during a recent restructure.

basaks:
@chandra2ga, Can you double check your first ovservation. When I test using this command:


>> mpirun -n 4 pyrate workflow -f tests/test_data/small_test/conf/pyrate_gamma_test.conf 


This file is created:
>> out/gamma/out/ref_pixel_150.941666654_-34.218333314_5_5_5_0.8_file.npy

This test also proves it's created and reused in subsequent runs:

>>https://github.com/GeoscienceAustralia/PyRate/blob/febc8837b60fb19183ceefaa07c64c34d3cd4562/tests/test_refpixel.py#L433

There are several reasons for the chosen filenames:

1. The whole reason you want the parameter names in the filename is that you don't overwrite that file, and can reuse that file if it's again on disc. In this situation if you provided (-1, -1) first, then changed to a (lat, lon), and then went back to (-1, -1), the saved refpixel file will not be overwritten and reused when you go back to the old param. You suggestion will overwrite file, and in this example, on the 3rd job, refpixel will have to be recomputed. ref_pixel.npy does not work and additional effort was spent to make this change.
2. These filenames are not for users to inspect, while still being able to do so. The don't have to be pretty. These are intermediate files to enhance usability. 
3. This is an extract string representation of the config parameters used.


Your second point, again the user does not have to really know which parameters are in the orbfit filename, but if the appropriate file is available on disc, they are reused if the parameters match. If you want to know which paramters goes in which sequence, you can inspect this line of code for the orbfit file names:

>> https://github.com/GeoscienceAustralia/PyRate/blob/febc8837b60fb19183ceefaa07c64c34d3cd4562/pyrate/configuration.py#L108

basaks:
@mcgarth multillooking for orbfit also requries effort.

Since independent method is trivially parallisable, we don't really need multilooking in this case. We have to multilook the file and then take median in this case. Why spend the effor in multilooking? What does it gain us? In memory multilooking is available due to network method, and can be very quickly implemented.

The network method on the other hand has more reason to use multilooking as we need all interferrograms in memory, and that step is not parallel.

Have a rethink the filenames based on my previous comments.

chandra2ga:
> > > out/gamma/out/ref_pixel_150.941666654_-34.218333314_5_5_5_0.8_file.npy

@basaks: Yeah, I noticed this file is created with default pixel values on config file. I might missed that.

> 1. The whole reason you want the parameter names in the filename is that you don't overwrite that file when parameters change on the next run, and can reuse that file if it's available on disc. In this situation if you provided (-1, -1) first, then changed to a (lat, lon), and then went back to (-1, -1), the saved refpixel file will not be overwritten and reused when you go back to the old param. You suggestion will overwrite file, and in this example, on the 3rd job, refpixel will have to be recomputed. ref_pixel.npy does not work and additional effort was spent to make this change.

As as per this program when we have default ref pixel location on config file a new refpixel file will create if not already generated on disc and use that file for next processing. In case of automatic pixel lat/lon selection it will produce a new file name ref*-1-1*.npy and if that file already present on disc, program will reuse that file for next processing step without further overwriting or recreating. But as per our new workflow, reference pixel step will work after performing all the correction steps, so in that case selection of automatic reference pixel might vary depending on additional correction steps included to the workflow. And if the file will not overwrite to the exiting ref*-1_-1*.npy during automatic pixel selection, it might choose an incorrect pixel location reading from the old existing file. 

In any case (-1,-1) on file not much convincing instead could be replaced by lat/lon as created with default one. This will discard any issue of overwriting then.

basaks:
@chandra2ga , That does not work as when the reference pixel file is not deterministic at the beginning of the file, you cannot reuse it. The reference pixel location when the steps and sequence of corrections (which are interdepepdent) will vary. So we cannot adopt such a strategy. In this case we have to drop the reuse of reference pixels idea.

PR #285, opened 2020-08-06
Linrate: Linear regression of cumulative displacement time series

mcgarth:
This PR implements new functionality to calculate the rate for each pixel as the linear regression of the cumulative displacement time series.

The functionality is based on the `scipy.stats.linregress` function: https://docs.scipy.org/doc/scipy-1.4.1/reference/generated/scipy.stats.linregress.html?highlight=linregress#scipy.stats.linregress

**Input:** cumulative displacement time series `tscuml`

**Output:**
- `linrate` = gradient of best fitting line
- `intercept` = y-intercept of best fitting line at t = 0
- `rsquared` = R^2 goodness of fit statistic [0 ->1]
- `error` = standard error of the best fitting line w.r.t the observations
- `samples` = number of observations used to constrain the best fit line

All five outputs are saved as `linear_*.tif` files after the `merge` step and optionally as `linear_*.npy` files if `savenpy = 1`. Additionally `linrate`, `rsquared` and `error` are saved as png and kml files.

The Linrate functionalty is executed as a non-optional part of the timeseries algorithm (it will be calculated if `tscal = 1`)

This PR implements unit tests at the pixel and array level (added to `test_timeseries.py`) and regression tests in `test_mpi_vs_multiprocess_vs_single_process.py`


pyrate/core/timeseries.py
```diff
@@ -302,6 +303,102 @@ def _solve_ts_lap(nvelpar, velflag, ifgv, mat_b, smorder, smfactor, sel, vcmt):
...
+    :return: error: Standard error of the linear regression
+    :rtype: float
+    :return: samples: Number of observations used in linear regression
+    :rtype: int
+    """
+
+    # test for equal length of input vectors
+    if len(y) != len(t):
+        raise TimeSeriesError("linear_rate_pixel: y and t are not equal length")
+
```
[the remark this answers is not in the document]
mcgarth: The point of the check is to do error checking on the input in case someone calls this function directly, outside of the context of the workflow. The raised exception gives a more contextual message than the `linregress` exception. How much time do we expect to save by removing this check on say, 1 Million function calls (i.e. a spatial array of 1000 x 1000)?

mcgarth:
> I reviewed this PR with a new data stack and here are my observations:
> 
> 1. The program produces and saved all the five expected outputs (linear_rate, error, intercept, samples, rsquared) and also saved optional linear_*.npy files to the disc.
> 2. Concern to results analysis between linear and stack rate. This needs further discussion understanding the performance analysis between two approaches.
> 3. At the moment linear_rate result shows higher deformation range limits compared to stack_rate.
>    Similarly for linear_error which shows a greater error limit as compare to stack_error. I could expect differences between both approaches but greater than one order of magnitude in case of linear rate and error calculation don't sound reasonable. This may need a further investigation on performance of algorithm.
> 
> But I approve PR as it generates all the required files as expected and identifying structures similar to stack calculation.

Thanks for the comments @chandra2ga. I think we can really do some performance evaluation of both `stack_rate` and `linear_rate` once we have some GPS data available for comparison

pyrate/core/timeseries.py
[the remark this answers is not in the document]
uniomni: Good discussion. Another point of view could be to use exception handling to catch and reraise the linregress exception and turn it into something more meaningful to the caller - and at the same time avoiding doing the same check 1 million times. Admittedly, whether this is a performance issue should be substantiated by profiling.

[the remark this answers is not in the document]
mcgarth: Thanks for commit 72aae9e @basaks , which I think is a good compromise. I have also used @uniomni's suggestion to re-raise the exception in the `linear_rate_pixel` function to give a more context specific exception message. Good outcome!

mcgarth:
> I've performed a full run-through PyRate using the mg/linrate branch and can confirm that all works as expected. I've used Camden RSAT2 data for this test.
> 
> I can also confirm that the linear_rate result is noiser compared to stack_rate result as mentioned by Chandra (though not that much different as in Chandra's test case. I also realise that the linear_rate result considers more pixels compared to the stack_rate result (see images below, colour scales are +/- 273 and +/-248, respectively). The difference may be related to the masking of pixels (r-value) and the handling of uncertainties of the input data (i.e. the cumulative displacements).
> 
> Future improvement:
> The scipy function 'linregress' is used to calculate the velocity + uncertainty for each pixel. I understand that 'linregress' does not support inputting uncertainties of the y-data (i.e. the cumulative displacements), not talking about temporal correlations using a full varcov matrix. You may want to add consideration of uncertainties and temporal correlations in the linrate algorithm to the list of tickets.
> 
> stack_rate.png:
> stack_rate
> 
> linear_rate.png:
> linear_rate

Thanks for this review @tfuhrmann . Your visual comparison gives me more confidence in the software - we are able to resolve known deformation signals. Interesting that `linear_rate` appears to resolve the deformation zones better than `stack_rate`. But yes there seems to be more noise in the `linear_rate` result. No masking has been applied in the algorithm; this could be done during post-processing by using the `linear_samples` or `linear_rsquared` layers as a mask. Your idea to add time series uncertainties to the linear regression is a good one. The only problem is we don't currently compute time series uncertainties!

pyrate/core/timeseries.py
```diff
@@ -302,6 +303,110 @@ def _solve_ts_lap(nvelpar, velflag, ifgv, mat_b, smorder, smfactor, sel, vcmt):
...
+    :return: rsquared: R-squared value of the linear regression
+    :rtype: float
+    :return: error: Standard error of the linear regression
+    :rtype: float
+    :return: samples: Number of observations used in linear regression
+    :rtype: int
+    """
+
+    while True:
+        try:
```
basaks, line 327: don't need the `while True` and `break`? just `try and except` are enough?

mcgarth: Removed. Thanks for the tip! I followed a recipe that had them in

pyrate/core/timeseries.py
```diff
@@ -302,6 +303,108 @@ def _solve_ts_lap(nvelpar, velflag, ifgv, mat_b, smorder, smfactor, sel, vcmt):
...
+
+    # Mask to exclude nan elements
+    mask = ~isnan(y)
+    # remove nan elements from both arrays
+    y = y[mask]    
+    try:
+        t = t[mask]
+    except IndexError:
+        raise TimeSeriesError("linear_rate_pixel: y and t are not equal length")
+
```
basaks, line 334: One more time, sorry! Given `len(y) == len(t)` prior to entering this function as checked in the calling function, is the `IndexError` a possibility?

mcgarth: yes, this is there in case someone calls this function directly, and independently of the PyRate workflow. In that situation the could inadvertently pass unequal length vectors.

basaks: There are no consumers for this though? It should probably be a protected function. Functions that do the heavy lifting/i.e., called billions of time should avoid such interruptions.

mcgarth: I see your point but don't think it should be enforced for this PR because I know that this idea is not already consistently applied across PyRate. However, I do think as a result of this discussion that we should set aside some time for a package-wide review of function exposure and protection and do something consistently across PyRate. I'll lodge a ticket in the backlog ;-)

PR #287, opened 2020-08-11
Stored ~isnan(y) in a variable for readability and possibly performance

uniomni:
Dipping my toes in the water


mcgarth:
Good suggestion, thanks!

PR #288, opened 2020-08-12
Sb/step4 refactor process

basaks:
1. all process/correct steps except maxvar and vcmt calcs are saved on disc and reused
2. `process` renamed to `correct`, all references and tests updated. `correct` does not compute `timeseries` or `stack`ing
3. new `timeseries` and `stack` commands.
4. max var mpi code simplified.
5. more mpi parallelisation in spatial and temporal filtering during aps correction. APS correction can be many X faster based on CPUs thrown at it. May be even >10x based on num of cpus.
6. Test suite extended to cover each of these steps (we now have a test_aps.py too).
7. (MG) Processing of coherence files by conv2tif and prepifg is now triggered by the presence of `cohfilelist` in the config file. If the list is present, multilooked/cropped coherence files are saved to disk, regardless of whether coherence masking is on or off (parameter `cohmask` in config file).
8. (MG) Deprecate redundant `tscal` config parameter, that was used previously to control time series execution. This is not needed anymore now there is a `timeseries` step invokable on the command line.
9. add unit test coverage for refpixel lat/lon to x/y conversion.
10. A common framework for using tiles for both mpi and multiprocessing methods in stack, timeseries and mst algorithms.


chandra2ga (approved):
I reviewed this PR with the latest commits and here are my observations:

1) The program working as per our current PyRate workflow. No "process" step exits anymore instead new refactor functions "correct", "timeseries" and "stack" steps working as per our expectation.
2) We need to remove name "Step 3: process" from the PyRate workflow while using command 'pyrate -h'. Instead we should update with the new function steps to 'pyrate -h'.
3) The output files from MST, VCM and Ref phase correction steps are currently generating from "correct" step, which is bit different than our existing latest workflow. But the team happy the with current status of processing (this is after discussion with Thomas and Matt), no need for further modification. I will update the PyRate workflow diagram accordingly.  
4) Running "timeseries" and "stack" steps are optional and user dependent. Both steps running independent to each other without any error or issues. 
5) Now "merge" step is also correctly using the output files available from previous steps (either stack or timeseries or both), and generating stack_rate and/or linear_rate as per files available from early steps. 
6) "conv2tif" and "prepfig" steps are using 'cohfilelist' path on the config file for processing (as mention in point-7)

Hence, I approved the PR. Good to include this PR for next release. Well done!

mcgarth (commented):
At the user level, I am happy with the changes in this PR and it implements the user functionality requested by the team. I have pushed a few commits on top of the branch that make some minor house-keeping changes. The biggest change in these commits is deprecating the `tscal` variable, which this PR makes redundant.

I did some numeric comparison of this PR against `develop`, by comparing summary statistics of output geotiffs (`gdalinfo -stats`) after running the full `workflow` against the unit test data. The results are documented in [this word doc](https://github.com/GeoscienceAustralia/PyRate/files/5100063/step4-refactor-process_review.docx). In summary, there are differences in the mean and std-dev statistics. The worst is a difference in the 6th decimal place (of a millimetre), which I think is acceptable given that the measurement precision of InSAR is no where near 0.000001 millimetres! However, an investigation in to why these differences are occurring for this PR is recommended please @basaks . 

I also ran a benchmark to test the improvements made by parallelising the spatial and temporal filters. For this benchmark on `gadi` I used 153 Mexico ifgs, 128Gb RAM and `mpirun -n 16 pyrate workflow ...`. With `apsest=0`, runtimes were 5m 33s and 5m 24s for `develop` and `sb/step4-refactor-process` respectively. With `apsest=1`, runtimes were 9m 40s and 8m 14s for `develop` and `sb/step4-refactor-process` respectively. This represents an improvement in efficiency of ~15%. Well done @basaks!

Minor comments:
- when running using MPI, every process writes every log message. Is there a way to suppress this, so that only the control process writes log messages?
- It is not obvious to the user that VCMT is saved to disk (it is buried in `correction.params`). Suggest saving the VCMT to a numpy array file `vcmt.npy`

pyrate/correct.py
```diff
@@ -150,18 +138,38 @@ def _update_params_with_tiles(params: dict) -> None:
...
+
+
+def load_params_from_disc(params: dict) -> dict:
+    params_file = Path(params[cf.OUT_DIR], 'correction.params')
+    if not params_file.exists():
+        raise FileNotFoundError("Params file not found on disc. Please run 'correct'")
+
+    with open(params_file, 'rb') as f:
+        params = cp.load(f)
+    return params
```
mcgarth, line 169: This function is only called in `pyrate.main` so I suggest moving it to `main.py` and make it an internal function

mcgarth: I can see this function `load_params_from_disc` has instead been deprecated

mcgarth:
The additional changes by @basaks have addressed all the issue I raised!
I re-did my numeric comparison tests for one run (`apsest=1`, `cohmask=1`) - the results were identical. So saving VCMT to disk as `vcmt.npy` indeed did not change a thing as suspected by @basaks. I am happy with the explanation for the changes at 6 d.p. level given by @basaks above, and accept them. They are only fractions of a millimetre in magnitude and won't trouble us

PR #289, opened 2020-08-12
Mg/linrate

uniomni:
Updated README.rst with instructions how to run the tests


uniomni:
Made input checks in more explicit. I think it reads better that way.

mcgarth:
I have cherry-picked the changes to README.rst and commited them to PyRate's develop branch: https://github.com/GeoscienceAustralia/PyRate/commit/fa33f542dc63de3967c529a15c83a1e946c24418

I have not included the explicit check for vector length included in this PR. But in the version of the function `linear_rate_pixel` merged in to develop, an IndexError is raised when we do the operation t = t[mask] and unequal vectors y and t were passed to the function. I catch that and raise a "TimeSeriesError" exception with a message saying that "y and t are not equal length". I think this should make it clear to the user that y and t should be equal. See https://github.com/GeoscienceAustralia/PyRate/commit/f99a038bbd216a8cbbaf898d5bc2612ed9009673

PR #290, opened 2020-08-17
Mg/plot

mcgarth:
PyRate does not save the zero epoch `tscuml_*.tif` to disk since it is a zero array and not worth saving. The only way to know the date of the missing zero epoch is to grab it from the geotiff metadata header for `linear_rate.tif`.

This PR changes the plotting script to grab the time series dates from the `linear_rate.tif` header. This fixes the zero-epoch issue in the existing script.

Other small changes:
- get the path variable from the command line (remove hard-coded path variable)
- use blue-white-red colour map (for consistency with quick look images)
- make tscuml the consistent naming convention for variables
- plot `linear_rate.tif` product rather than `stack_rate.tif`
- use datetime variable for x-axis (rather than ordinal values)


chandra2ga (approved):
Thanks Matt for the changes. It's working fine. Few additional changes we need to include in this script.
1) rasterio module needs to include in PyRateVenv before running the script. Otherwsie shows error
2) We need to change the datetime in tslider as well. 
3) For changing lat/lon to the vel map need a bit of work. I noticed it's affecting my code when I moved to lat/lon map from radar co-ordinate.
4) I added a commit for x-axis plot as datetime (early code the scatter plot and line fit are not overlay correctly on the same window while plotting)

Rest all looks alright.

PR #293, opened 2020-08-21
Cko/plot_time_series

chandra2ga:
This is an updated version of plot_time_series script. The modification applied in this PR are as follows :

1) PyRate linear fit results included in the time-series 
2) Different color plots for PyRate linear and model fits
3) Applied changes to the datetime ordinal values on x-axis of time-slider 
4) Resolved issue of displaying cumulative displacement for different steps.


mcgarth (approved):
Thanks for this contribution @chandra2ga - it's a fantastic tool for us and future users of PyRate to interact with output tif products.
I have added a few commits that change the following:
- remove duplicated code (establish two new functions)
- fixed time series plotting when there are NaNs in the tscuml data (e.g. unit test data set)
- add grey background to map images
- add licence boilerplate

PR #294, opened 2020-08-21
Sb/timeseries uses multiprocsesing via tiles

basaks:
Common framework for using tiles via both `mpi` and `multiprocessing`.
Stack, timeseries and mst use this commong framework
Addresses some comments in the orignal PR. https://github.com/GeoscienceAustralia/PyRate/pull/288


mcgarth (approved):
Some good changes here. The general framework for parallelism will make future development of new functionality much easier!

PR #295, opened 2020-08-25
Release 0.5.0

mcgarth:
This PR completes the refactor of the PyRate workflow to enable easier analysis of correction data and quicker re-computation with different parameter settings

**Added**
- New functionality "``linear_rate``" to calculate linear regression of
  cumulative displacement time series for every pixel as part of the ``timeseries`` step.
- Script for plotting ``timeseries`` and ``linear_rate`` output geotiff products using Matplotlib.
  To use, additional dependencies listed in ``requirements-plot.txt`` are required.
- Correction data (except ``maxvar`` and ``vcmt``) applied to the ifg data is saved to disk
  and re-used on subsequent repeat runs. Corrections are only re-calculated if config
  parameters change between runs.
- MPI parallelisation of APS spatio-temporal filter correction.
- Unit test coverage for refpixel lat/lon to x/y conversion and ``aps`` module.

**Fixed**
- Re-enable ``ifglksx`` and ``ifglksy`` to be different values, resulting in different
  resolutions in x and y dimensions in multi-looked interferograms.
- Re-enable ``orbfitlksx`` and ``orbfitlksy`` to be different values, resulting in different
  resolutions in x and y dimensions during network orbit correction.
- Screen messages from main process only during MPI runs.

**Changed**
- ``process`` step has been renamed ``correct``. Stacking and timeseries have been removed from
  this step and are now invoked by separate ``timeseries`` and ``stack`` command line options.
- Processing of coherence files by ``conv2tif`` and ``prepifg`` is now triggered by the presence
  of ``cohfilelist`` in the config file. If the list is present, multilooked/cropped coherence
  files are saved to disk, regardless of whether ``cohmask`` is 0 or 1.
- Parallelisation capability is refactored - MPI and multiprocessing both now use a common
  tiling framework for ``stack``, ``timeseries`` and ``mst`` algorithms.
- Introduced a simplified and standardised file naming format for files produced by the 
  ``prepifg`` step. Info previously saved in the filename (e.g. ``ifglksx``, ``ifglksy``, ``ifgcropopt``)
  is now saved to the geotiff header instead.

**Removed**
- Redundant ``tscal`` config parameter was deprecated - not needed now there is a ``timeseries``
  step invokable on the command line.
- `Pillow`, `cython` and `glob2` dependencies - not used in main workflow code
- Deprecated function `pyrate.prepifg_helper.prepare_ifgs`, which is no longer needed.

PR #297, opened 2020-08-31
Sb/disable network orbfit and offset user params

basaks:
1. remove orbitoffset param and prodide defaults. deafults are `offset=false` for independent method and `offset=true` for network method.
2. use orbfit independent method as fallback until network method is fixed


mcgarth (approved):
I think the changes in this PR adequately address the actions raised in https://github.com/GeoscienceAustralia/PyRate/pull/295

PR #300, opened 2020-09-06
Sb/remove prepare ifgs

basaks:


mcgarth (approved):
Thanks for doing this @basaks!

PR #302, opened 2020-09-09
Tf/baseline geometry - Part 1 & Part 2

tfuhrmann:
First PR related to the implementation of baseline and geometry information into PyRate. The new functionality reads additional datasets from GAMMA interferogram stacks, i.e.:
- additional metadata from the slc/mli.par files,
- information on the spatial baseline vector of each interferogam as given in the *base.par files,
- the GAMMA lookup table file which is used to relate DEM rows/cols (and their respective lon/lat values) to radar-coded rows and cols (i.e. the azimuth/range coordinate system). Note that the binary lookup table file is transformed into numpy arrays for azimuth and range in PyRate.

I've added some new files/additional parameters to the test data set, and also needed to add code to the PyRate tests/*.py-files to get the test suite up and running again. All tests are now running successfully using pytest (except for some of the known Legacy test failures in Gadi) and are also passed in Travis . Note that additional tests for the new data files and functionalities need to be added. Also, the implementation of the functionality is not finished (e.g. I want to add code to use the "initial baseline" in case the "precision baseline" is not available in the base.par files). I have raised this PR anyway now, since I wanted to get up-to-date with the latest release.


pyrate/core/config.py
```diff
@@ -502,6 +515,72 @@ def coherence_paths_for(path: str, params: dict, tif=False) -> str:
...
+    """
+
+    dest_mlooked_ifgs = [mlooked_path(os.path.basename(q).split('.')[0] + '_'
+                                      + os.path.basename(q).split('.')[1] +
+                                      '.tif', looks=looks, crop_opt=crop)
+                         for q in base_paths]
+
+    return [os.path.join(params[OUT_DIR], p) for p in dest_mlooked_ifgs]
+
+
```
mcgarth, line 583: these two functions were deprecated in release 0.5.0. They don't exist in the develop branch

pyrate/core/config.py
```diff
@@ -639,6 +718,22 @@ def coherence_paths_for(path: str, params: dict, tif=False) -> str:
...
+}
+"""dict: basic validation functions for baseline parameters."""
+
+_LOOKUPTABLE_VALIDATION = {
+    LT_FILE: (
+        lambda a: a is not None and not os.path.exists(a),
+        f"'{LT_FILE}': if file is provided it must exist."
+    ),
+}
+"""dict: basic validation functions for baseline parameters."""
```
mcgarth, line 735: `baseline` should be `lookuptable`

pyrate/core/gamma.py
```diff
@@ -38,6 +38,20 @@
...
+GAMMA_RANGE_N = 'range_samples'
+GAMMA_AZIMUTH_PIX = 'azimuth_pixel_spacing'
+GAMMA_AZIMUTH_N = 'azimuth_lines'
+GAMMA_AZIMUTH_LOOKS = 'azimuth_looks'
+GAMMA_PRF = 'prf'
+GAMMA_NEAR_RANGE = 'near_range_slc'
+GAMMA_SAR_EARTH = 'sar_to_earth_center'
+GAMMA_SEMI_MAJOR_AXIS = 'earth_semi_major_axis'
+GAMMA_SEMI_MINOR_AXIS = 'earth_semi_minor_axis'
+# to do add option to use initial baseline if precision baseline was not calculated in GAMMA
```
mcgarth, line 52: as a tip, use "TODO" instead of "to do". TODO is recognised as a flag in `pylint` code checking and will be collated in to a list of TODO's

pyrate/core/gamma.py
```diff
@@ -136,14 +207,39 @@ def parse_dem_header(path):
     return subset
 
 
+def parse_baseline_header(path):
+    """
+    Returns dictionary of DEM metadata required for PyRate
+
+    :param str path: `Full path to Gamma *dem.par file`
```
mcgarth, line 214: fix up doco for baseline, not DEM

pyrate/core/config.py
tfuhrmann: Thanks for clarifying. That was one of the merge conflicts which I wasn't sure about. The new functionality doesn't need these functions either. They can hence be deleted.

pyrate/core/gamma.py
```diff
@@ -188,14 +284,108 @@ def combine_headers(hdr0, hdr1, dem_hdr):
...
         msg = "Wavelength mismatch, check both header files for %s & %s"
         raise GammaException(msg % args)
+
+    # use parameter of first image (as done by GAMMA during interferometric processing)
+    heading_ang = hdr0[ifc.PYRATE_HEADING_DEGREES]
+    if np.isclose(heading_ang, hdr1[ifc.PYRATE_HEADING_DEGREES], atol=1e-1):
+        chdr[ifc.PYRATE_HEADING_DEGREES] = heading_ang
+    else:
+        args = (chdr[ifc.FIRST_DATE], chdr[ifc.SECOND_DATE])
+        msg = "Satellite heading angles differ by more than 1e-1"
```
mcgarth, line 294: I wonder whether tiis worth making these error messages a bit more "human readable", for instance:

```
msg = "Satellite heading angles differ by more than 1 degree"

msg = "SAR to Earth Center distance differs by more than 1000 metres"
```

pyrate/core/gamma.py
```diff
@@ -249,7 +443,11 @@ def gamma_header(ifg_file_path, params):
     """
     dem_hdr_path = params[cf.DEM_HEADER_FILE]
     header_paths = get_header_paths(ifg_file_path, params[cf.HDR_FILE_LIST])
-    combined_headers = manage_headers(dem_hdr_path, header_paths)
+    if len(header_paths) == 2:
+        baseline_path = cf.baseline_paths_for(ifg_file_path, params)
+    else:
+        baseline_path = None  # don't read baseline files for DEM
+    combined_headers = manage_headers(dem_hdr_path, header_paths, baseline_path)
```
mcgarth, line 450: Just an observation, that I think this function call makes baseline files a mandatory input for a GAMMA-based PyRate run...
baseline_path could be made an optional arg to `manage_headers`, defaulting to 'None' ...?

pyrate/core/ifgconstants.py
```diff
@@ -34,6 +34,23 @@
 PYRATE_TIME_SPAN = 'TIME_SPAN_YEAR'
 PYRATE_WAVELENGTH_METRES = 'WAVELENGTH_METRES'
 PYRATE_INCIDENCE_DEGREES = 'INCIDENCE_DEGREES'
+PYRATE_HEADING_DEGREES = 'HEADING_DEGREES'
+PYRATE_RANGE_PIX_METRES = 'RANGE_PIX_METRES'
+PYRATE_RANGE_N = 'RANGE_N'
+PYRATE_AZIMUTH_PIX_METRES = 'AZIMUTH_PIX_METRES'
+PYRATE_AZIMUTH_N = 'AZIMUTH_N'
+PYRATE_AZIMUTH_LOOKS = 'AZIMUTH_LOOKS'
```
mcgarth, line 42: Is it worth adding RANGE_LOOKS for completeness of information, even if it is not used for a calculation?

pyrate/core/shared.py
```diff
@@ -68,6 +68,7 @@
 class InputTypes(Enum):
     IFG = 'ifg'
     COH = 'coh'
+    BAS = 'bas'
```
mcgarth, line 71: I understand why you choose "bas" (to conform to the other three character abbreviations like "ifg", "coh", "dem"), but I think "base" might be a bit more readable in the code, and more instantly understandable. There is no real reason to keep these 3 characters, so 4 characters is also fine

pyrate/core/shared.py
```diff
@@ -717,6 +718,14 @@ def _is_coherence(hdr):
            (hdr[ifc.INPUT_TYPE] == InputTypes.COH if ifc.INPUT_TYPE in hdr else False)
 
 
+def _is_baseline(hdr):
+    """
+    Convenience function to determine if file is interferogram
```
mcgarth, line 723: *file is baseline

pyrate/core/config.py
tfuhrmann: see https://github.com/GeoscienceAustralia/PyRate/pull/302/commits/fa2a65850f4aec5bfc26712c9339f19b6edc51ee

pyrate/core/config.py
tfuhrmann: see https://github.com/GeoscienceAustralia/PyRate/pull/302/commits/fa2a65850f4aec5bfc26712c9339f19b6edc51ee

pyrate/core/gamma.py
tfuhrmann: see https://github.com/GeoscienceAustralia/PyRate/pull/302/commits/fa2a65850f4aec5bfc26712c9339f19b6edc51ee

pyrate/core/gamma.py
tfuhrmann: see https://github.com/GeoscienceAustralia/PyRate/pull/302/commits/fa2a65850f4aec5bfc26712c9339f19b6edc51ee

pyrate/core/gamma.py
tfuhrmann: see https://github.com/GeoscienceAustralia/PyRate/pull/302/commits/fa2a65850f4aec5bfc26712c9339f19b6edc51ee

pyrate/core/ifgconstants.py
tfuhrmann: see https://github.com/GeoscienceAustralia/PyRate/pull/302/commits/fa2a65850f4aec5bfc26712c9339f19b6edc51ee

pyrate/core/shared.py
tfuhrmann: see https://github.com/GeoscienceAustralia/PyRate/pull/302/commits/fa2a65850f4aec5bfc26712c9339f19b6edc51ee

pyrate/core/gamma.py
tfuhrmann: see https://github.com/GeoscienceAustralia/PyRate/pull/302/commits/fa2a65850f4aec5bfc26712c9339f19b6edc51ee

pyrate/core/dem_error.py
```diff
@@ -0,0 +1,307 @@
...
+    a = float(ifg.meta_data[ifc.PYRATE_SEMI_MAJOR_AXIS_METRES])
+    b = float(ifg.meta_data[ifc.PYRATE_SEMI_MINOR_AXIS_METRES])
+    se = float(ifg.meta_data[ifc.PYRATE_SAR_EARTH_METRES])
+    near_range = float(ifg.meta_data[ifc.PYRATE_NEAR_RANGE_METRES])
+    rps = float(ifg.meta_data[ifc.PYRATE_RANGE_PIX_METRES])
+    heading = float(ifg.meta_data[ifc.PYRATE_HEADING_DEGREES])
+
+    # convert angles to radians
+    lon = lon / 180 * np.pi
+    lat = lat / 180 * np.pi
```
mcgarth, line 118: please use numpy functions wherever possible, e.g. https://numpy.org/doc/stable/reference/generated/numpy.radians.html

pyrate/core/dem_error.py
```diff
@@ -0,0 +1,307 @@
...
+    transform = ifg.dataset.GetGeoTransform()
+    # number of rows and columns in dataset
+    nrows, ncols = ifg.shape
+    lon = np.zeros((nrows, ncols))  # pre-allocate 2D numpy array
+    lat = np.zeros((nrows, ncols))  # pre-allocate 2D numpy array
+    for i in range(0, nrows): # rows are y-direction
+       for j in range(0, ncols): # cols are x-direction
+           lon_ij, lat_ij = convert_pixel_value_to_geographic_coordinate(j, i, transform)
+           lon[i, j] = lon_ij
+           lat[i, j] = lat_ij
```
mcgarth, line 79: are these lines necessary? why not just do:
`lon[i, j], lat[i, j] = convert_pixel_value_to_geographic_coordinate(j, i, transform)`

pyrate/core/dem_error.py
```diff
@@ -0,0 +1,307 @@
...
+    """
+    Function to get radar coordinates for each pixel in the multi-looked interferogram dataset.
+    Radar coordinates are identical for each interferogram in the stack.
+    """
+    # lookup table file:
+    lookup_table = params[cf.LT_FILE]
+    # PyRate IFG multi-looking factors
+    ifglksx = params[cf.IFG_LKSX]
+    ifglksy = params[cf.IFG_LKSY]
+    print(lookup_table)
```
mcgarth, line 94: Is this just for testing? Use a `log.debug` statement instead of print if you want it to persevere in the code

pyrate/core/dem_error.py
```diff
@@ -0,0 +1,307 @@
...
+    # convert angles to radians
+    lon = lon / 180 * np.pi
+    lat = lat / 180 * np.pi
+    heading = heading / 180 * np.pi
+
+    # Earth radius at given latitude
+    re = np.sqrt(np.divide(np.square(a**2 * np.cos(lat)) + np.square(b**2 * np.sin(lat)), \
+                           np.square(a * np.cos(lat)) + np.square(b * np.sin(lat))))
+
+    # range measurement at pixel ij -> this requires the mli.par as input since rps would be different in the slc.par
```
mcgarth, line 125: this comment is specific to Gamma input. I guess in this module we are assuming that input is not Gamma specific anymore? In the future the input to this module may have come from software other than Gamma

pyrate/core/dem_error.py
```diff
@@ -0,0 +1,307 @@
...
+
+    # look angle at pixel ij -> law of cosines in "satellite - Earth centre - ground pixel" triangle
+    # see e.g. Section 2 in https://www.cs.uaf.edu/~olawlor/ref/asf/sar_equations_2006_08_17.pdf
+    look_angle = np.arccos(np.divide(se**2 + np.square(range) - np.square(re), 2 * se * range))
+
+    # incidence angle at pixel ij -> law of cosines in "satellite - Earth centre - ground pixel" triangle
+    # see e.g. Section 2 in https://www.cs.uaf.edu/~olawlor/ref/asf/sar_equations_2006_08_17.pdf
+    incidence_angle = np.pi - np.arccos(np.divide(np.square(range) + np.square(re) - se**2, 2 * np.multiply(range, re)))
+    # for validation with GAMMA output only
+    incidence_angle_gamma = np.pi / 2 - incidence_angle
```
mcgarth, line 136: move this out of prod code and in to a test for validating against Gamma output

pyrate/core/dem_error.py
```diff
@@ -0,0 +1,307 @@
...
+    # maximum differences to the GAMMA-derived local incidence angles for Sentinel-1 test data are within +/-0.1 deg
+
+    # local azimuth angle at pixel ij using constant satellite heading angle and spherical approximations
+    epsilon = np.pi - look_angle - (np.pi - incidence_angle) # angle at the Earth's center between se and re
+    # azimuth of satellite look vector(for right-looking SAR)
+    sat_azi = heading + np.pi / 2
```
mcgarth, line 142: as discussed, read this from the Gamma header file?

pyrate/core/dem_error.py
```diff
@@ -0,0 +1,307 @@
...
+    # the following equations are adapted from Section 4.4 (page 4-16) in EARTH-REFERENCED AIRCRAFT NAVIGATION AND
+    # SURVEILLANCE ANALYSIS (https://ntlrepository.blob.core.windows.net/lib/59000/59300/59358/DOT-VNTSC-FAA-16-12.pdf)
```
mcgarth, line 144: great to include these references as comments!

pyrate/core/dem_error.py
```diff
@@ -0,0 +1,307 @@
...
+    # the above code could be improved by calculating satellite positions for each azimuth row
+    # from orbital state vectors given in .par file, using the following workflow:
+    # 1. read orbital state vectors and start/stop times from mli.par
+    # 2. for each pixel get the corresponding radar row (from matrix az)
+    # 3. get the corresponding radar time for that row (using linear interpolation)
+    # 4. calculate the satellite XYZ position for that time by interpolating the time and velocity state vectors
+
+    # calc azimuth angle using Vincenty's equations
+    azimuth_angle = np.empty(lat.shape) # pre-allocate 2D numpy array
+    azimuth_angle[:] = np.nan
```
mcgarth, line 158: I think you can achieve the intent of these two lines in one (but I haven't tested my assumption):
`azimuth_angle = np.empty(lat.shape) * np.nan`

pyrate/core/dem_error.py
```diff
@@ -0,0 +1,307 @@
...
+    interferogram dataset. Coordinates are identical for each interferogram in the stack.
+    """
+    # assume all interferograms have same projection and will share the same transform
+    transform = ifg.dataset.GetGeoTransform()
+    # number of rows and columns in dataset
+    nrows, ncols = ifg.shape
+    lon = np.zeros((nrows, ncols))  # pre-allocate 2D numpy array
+    lat = np.zeros((nrows, ncols))  # pre-allocate 2D numpy array
+    for i in range(0, nrows): # rows are y-direction
+       for j in range(0, ncols): # cols are x-direction
```
mcgarth, line 76: indent here is only 3 spaces

pyrate/core/dem_error.py
```diff
@@ -0,0 +1,307 @@
...
+    lon = lon / 180 * np.pi
+    lat = lat / 180 * np.pi
+    heading = heading / 180 * np.pi
+
+    # Earth radius at given latitude
+    re = np.sqrt(np.divide(np.square(a**2 * np.cos(lat)) + np.square(b**2 * np.sin(lat)), \
+                           np.square(a * np.cos(lat)) + np.square(b * np.sin(lat))))
+
+    # range measurement at pixel ij -> this requires the mli.par as input since rps would be different in the slc.par
+    range = near_range + rps*rg
```
mcgarth, line 126: I think in general we want to add spaces between mathematical operators to improve human readability of the code:
`range = near_range + rps * rg`

pyrate/core/dem_error.py
```diff
@@ -0,0 +1,307 @@
...
+            azimuth_angle[ix, iy] = az12
+    np.reshape(azimuth_angle, lat.shape)
+
+    # old code: spherical azimuth angle calculation
+    #azimuth_angle2 = np.arccos(np.divide(np.multiply(np.sin(sat_lat), np.cos(lat)) - \
+    #                                np.multiply(np.multiply(np.cos(sat_lat), np.sin(lat)), np.cos(sat_lon - lon)), \
+    #                                np.sin(epsilon)))
+    #azimuth_angle_diff = azimuth_angle2 - azimuth_angle
+    #print(np.nanmin(azimuth_angle_diff), np.nanmax(azimuth_angle_diff))
+    # the difference between Vincety's azimuth calculation and the spherical approximation is ~0.001 radians
```
mcgarth: delete this if it is redundant. potentially you could use it in a test for validation of the preferred method?

pyrate/core/dem_error.py
```diff
@@ -0,0 +1,307 @@
...
+
+    # for validation with GAMMA output only
+    azimuth_angle_gamma = -(azimuth_angle - np.pi / 2) # local heading towards satellite
```
mcgarth, line 174: as before, move to a unit test in the test suite.

pyrate/core/dem_error.py
```diff
@@ -0,0 +1,307 @@
...
+    # maximum differences to the GAMMA-derived local azimuth angles for Sentinel-1 test data are within +/-0.5 deg
+    # this could be improved by using orbital state vectors to calculate that satellite positions (see above comment)
+
+    # save angles to npy arrays
+    look_angle_file = os.path.join(params[cf.OUT_DIR],'look_angle.npy')
+    np.save(file=look_angle_file, arr=look_angle)
```
mcgarth: do we need *.npy and *.tif by default? We have a user variable `savenpy` that can trigger the *npy save if it is really wanted

pyrate/core/dem_error.py
```diff
@@ -0,0 +1,307 @@
...
+    look_angle_file = os.path.join(params[cf.OUT_DIR], 'look_angle.tif')
+    shared.write_output_geotiff(md, gt, wkt, look_angle, look_angle_file, np.nan)
+    incidence_angle_file = os.path.join(params[cf.OUT_DIR], 'incidence_angle.tif')
+    shared.write_output_geotiff(md, gt, wkt, incidence_angle, incidence_angle_file, np.nan)
+    azimuth_angle_file = os.path.join(params[cf.OUT_DIR], 'azimuth_angle.tif')
+    shared.write_output_geotiff(md, gt, wkt, azimuth_angle, azimuth_angle_file, np.nan)
+    incidence_angle_gamma_file = os.path.join(params[cf.OUT_DIR], 'incidence_angle_gamma.tif')
+    shared.write_output_geotiff(md, gt, wkt, incidence_angle_gamma, incidence_angle_gamma_file, np.nan)
+    azimuth_angle_gamma_file = os.path.join(params[cf.OUT_DIR], 'azimuth_angle_gamma.tif')
+    shared.write_output_geotiff(md, gt, wkt, azimuth_angle_gamma, azimuth_angle_gamma_file, np.nan)
```
mcgarth, line 194: not needed for prod code. suggest delete

pyrate/core/dem_error.py
```diff
@@ -0,0 +1,307 @@
...
+        ifg.open()
+
+    # read relevant metadata from IFG
+    prf = float(ifg.meta_data[ifc.PYRATE_PRF_HERTZ])
+    az_looks = int(ifg.meta_data[ifc.PYRATE_AZIMUTH_LOOKS])
+    az_n = int(ifg.meta_data[ifc.PYRATE_AZIMUTH_N])
+    #base_T = float(ifg.meta_data[ifc.PYRATE_BASELINE_T])
+    base_C = float(ifg.meta_data[ifc.PYRATE_BASELINE_C])
+    base_N = float(ifg.meta_data[ifc.PYRATE_BASELINE_N])
+    #baserate_T = float(ifg.meta_data[ifc.PYRATE_BASELINE_RATE_T])
```
mcgarth, line 214: delete T components from here if they'll never be used

pyrate/core/dem_error.py
```diff
@@ -0,0 +1,307 @@
...
+
+    # save perpendicular baseline to npy array
+    bperp_on_disc = MultiplePaths.bperp_path(ifg.data_path, params)
+    np.save(file=bperp_on_disc, arr=bperp)
+
+    # save bperp to geotiff (temporary for visualisation)
+    gt, md, wkt = shared.get_geotiff_header_info(ifg.data_path)
+    md[ifc.EPOCH_DATE] = None # needs to have a value in write_output_geotiff
+    bperp_tif_on_disc = MultiplePaths.bperp_tif_path(ifg.data_path, params)
+    shared.write_output_geotiff(md, gt, wkt, bperp, str(bperp_tif_on_disc), np.nan)
```
mcgarth, line 235: As before, tif be default. npy only if `savenpy = 1`

pyrate/core/dem_error.py
```diff
@@ -0,0 +1,307 @@
...
+    
+    return bperp
+
+
+def vincinv(lat1, lon1, lat2, lon2, semimaj, semimin):
+    """
+    Vincenty's Inverse Formula, adapted from GeodePy
```
mcgarth: give a persistent web address for where you grabbed the code, e.g. https://github.com/GeoscienceAustralia/GeodePy/blob/master/geodepy/geodesy.py

pyrate/core/dem_error.py
```diff
@@ -0,0 +1,307 @@
...
+                                 - sin(u1)*cos(u2)*cos(lon)))
+    if azimuth1to2 < 0:
+        azimuth1to2 = azimuth1to2 + 2*pi
+
+    return round(azimuth1to2, 9)
+
+
+def dem_error_calc_wrapper(params: dict) -> None:
+    """
+    MPI wrapper for orbital fit correction
```
mcgarth, line 303: fix up doco

pyrate/core/dem_error.py
```diff
@@ -43,18 +43,22 @@ def remove_dem_error(ifgs: List, params: dict) -> None:
...
-    look_angle = calc_local_geometry(ifg0, ifg0_path, az, rg, lon, lat, params)
-    # loop over all Ifgs in stack
-    ifgs = [shared.Ifg(p) for p in ifg_paths] if isinstance(ifgs[0], str) else ifgs
-    process_ifgs = mpiops.array_split(ifgs)
-    for ifg in process_ifgs:
-        bperp = calc_local_baseline(ifg, az, look_angle, params)
-    # todo: add code to invert for DEM error using the per-pixel baseline values and the unwrapped phase
+    # not currently implemented for ROIPAC data which breaks some tests
+    # if statement can be deleted once ROIPAC is deprecated from PyRate
+    if not ifg0.meta_data[ifc.PYRATE_INSAR_PROCESSOR] == 'ROIPAC':
```
mcgarth, line 48: Just a comment. By the time the code progresses to this module, it should be input-format agnostic. Meaning that preprocessing steps have grabbed what is needed from Gamma, ISCE, SNAP formats and converted it to a standard "Pyrate" format that this module expects.

pyrate/core/dem_error.py
```diff
@@ -43,18 +43,22 @@ def remove_dem_error(ifgs: List, params: dict) -> None:
...
+      lon, lat = get_lonlat_coords(ifg0)
+      # calculate per-pixel radar coordinates
+      az, rg = get_radar_coords(ifg0, params)
+      # calculate per-pixel look angle (also calculates and saves indcidence and azimuth angles)
+      look_angle = calc_local_geometry(ifg0, ifg0_path, az, rg, lon, lat, params)
+      # loop over all Ifgs in stack
+      ifgs = [shared.Ifg(p) for p in ifg_paths] if isinstance(ifgs[0], str) else ifgs
+      process_ifgs = mpiops.array_split(ifgs)
+      for ifg in process_ifgs:
+          bperp = calc_local_baseline(ifg, az, look_angle, params)
```
mcgarth, line 60: So this is implementing parallelism by sending each ifg to a process to be done independently? What's the performance like for full-size data? Is it worth tiling-up the per pixel calculations?

mcgarth:
I think it's worth us having a discussion on when these new elements are computed. My thinking is that baseline, incidence and azimuth images are done during `prepifg`. They are not a 'correction'. They are tools used in correction(s) in the `correct` step. Furthermore, when we implement reprojection (to pseudo-vertical or -horizontal) I think it makes sense to do that in `prepifg` also, so we would need the azimuth and incidence images in that step.
Clearly we need a `dem_error.py` module that is called in the `correct` step. But I am not sure the current contents of `dem_error.py` is best housed there.
Open to hearing other peoples views though @tfuhrmann @adeane-ga @chandra2ga @basaks

tfuhrmann:
> I think it's worth us having a discussion on when these new elements are computed. My thinking is that baseline, incidence and azimuth images are done during `prepifg`. They are not a 'correction'. They are tools used in correction(s) in the `correct` step. Furthermore, when we implement reprojection (to pseudo-vertical or -horizontal) I think it makes sense to do that in `prepifg` also, so we would need the azimuth and incidence images in that step.
> Clearly we need a `dem_error.py` module that is called in the `correct` step. But I am not sure the current contents of `dem_error.py` is best housed there.
> Open to hearing other peoples views though @tfuhrmann @adeane-ga @chandra2ga @basaks

Thanks Matt for all your comments. I'll work on those today along with fixing the tests which are currently not passed in Travis.
On your comment: I agree that the tools currently implemented in `dem_error.py` are not the correction as such but only the geometry/baseline input which will be required to calculate the correction. So, I'm happy to move those parts to the end of `prepifg`. The only reason I've put it in the correct step was that all input can be calculated 'on-the-fly' and doesn't need to be saved to disc necessarily (currently only output for testing). But yes, if we are going to make use of look/incidence/azimuth angle files in the future and want to output them anyway, it makes sense to move these parts to the end of `perpifg` and then read the look_angle.tif to calculate the DEM error correction.

pyrate/core/dem_error.py
tfuhrmann: Already fixed in one of the later commits:
Vincenty's Inverse Formula, adapted from GeodePy function vincinv
(see https://github.com/GeoscienceAustralia/GeodePy/blob/master/geodepy/geodesy.py)

pyrate/core/dem_error.py
tfuhrmann: I have commented that out in a later commit, since it was output for testing only. In general, I'm not sure whether we want to output .npy or .tif files. I'm tending towards tif.

pyrate/core/dem_error.py
tfuhrmann: Was just for testing. Deleted.

pyrate/core/dem_error.py
tfuhrmann: Removed comment on mli.par and added a comment to `gamma.py`.

pyrate/core/dem_error.py
tfuhrmann: Done. Didn't know that this function exists.

pyrate/core/dem_error.py
tfuhrmann: Done.

pyrate/core/dem_error.py
tfuhrmann: To be done once we've got a new test data set with the required GAMMA outputs. Commented out for now with a todo note.

pyrate/core/dem_error.py
tfuhrmann: That's right. Changed as suggested. This might however be slightly slower (see discussion here: https://stackoverflow.com/questions/1704823/create-numpy-matrix-filled-with-nans: "I like this one: a = numpy.empty((3, 3,)) * numpy.nan. It timed faster than fill but slower than the assignment method, but it is a oneliner!!")

pyrate/core/dem_error.py
tfuhrmann: To be done once a test data set with all relevant input for the baseline/geometry functionality is implemented. For now I've commented the calculation and output of *_angle_gamma files out and made a todo comment on moving this into a test.

pyrate/core/dem_error.py
tfuhrmann: Done. Also changed `range` to `range_ij` to avoid confusion with the Python `range()` function.

pyrate/core/dem_error.py
tfuhrmann: Good point. Added a todo comment to move this to tests.

pyrate/core/dem_error.py
tfuhrmann: Deleted.

pyrate/core/dem_error.py
tfuhrmann: Done.

pyrate/core/dem_error.py
tfuhrmann: Deleted (was for testing only).

pyrate/core/dem_error.py
tfuhrmann: Agree. I only needed to add this line due to some ROIPAC tests failing.

pyrate/core/dem_error.py
tfuhrmann: Changed to 4 spaces.

pyrate/core/dem_error.py
tfuhrmann: Done.

pyrate/core/dem_error.py
tfuhrmann: Implemented reading of parameter `azimuth_angle` from GAMMA mli.par file and using its value (i.e. +90 deg or pi/2 for right-looking SAR sensors) instead of restricting the equation to right-looking SAR sensors.

pyrate/core/dem_error.py
tfuhrmann: Commented out and added a todo comment.

pyrate/core/dem_error.py
tfuhrmann: To be tested once I've implemented the actual DEM error correction.

mcgarth:
I have added commit 8b9ba31 to make the calculation of bperp files and geometry files optional, triggered by presence of `basefilelist` and `ltfile` parameters in the config file

Once tests are passing, I am happy for this PR to be merged in to `develop` as a work in progress. Feedback below can be incorporated in to the next PR.

I have tested with the unit test data. When `basefilelist` and `ltfile` are in the config file, I find that the baseline metadata (six entries) are added to headers in ifg and coh geotiff files from both `conv2tif` and `prepifg`. Even without `basefilelist` or `ltfile`, additional metadata from the slc/mli.par files is saved in to the ifg/coh geotiff headers. During `correct` step, look_angle.tif, incidence_angle.tif, azimuth_angle.tif and dem_error/*bperp.tif are generated. It looks like the first three contain angles in units of radians (would degrees be a better choice for quick human interpretation?). Some thought needs to go in to the metadata to be included in the headers of these files; currently they're labelled as `DATA_TYPE=MULTILOOKED_IFG`, which isn't accurate. A DATA_UNITS entry could also be useful.

The function `pyrate.core.shared.read_lookup_table` is specific to a Gamma-format flat-binary FCOMPLEX lookup table file. I therefore think it would be better housed in `pyrate.core.gamma` and executed during `prepifg`.

tfuhrmann:
> I have added commit [8b9ba31](https://github.com/GeoscienceAustralia/PyRate/commit/8b9ba31b3f54780617f001da696cbb1dac04130f) to make the calculation of bperp files and geometry files optional, triggered by presence of `basefilelist` and `ltfile` parameters in the config file
> 
> Once tests are passing, I am happy for this PR to be merged in to `develop` as a work in progress. Feedback below can be incorporated in to the next PR.
> 
> I have tested with the unit test data. When `basefilelist` and `ltfile` are in the config file, I find that the baseline metadata (six entries) are added to headers in ifg and coh geotiff files from both `conv2tif` and `prepifg`. Even without `basefilelist` or `ltfile`, additional metadata from the slc/mli.par files is saved in to the ifg/coh geotiff headers. During `correct` step, look_angle.tif, incidence_angle.tif, azimuth_angle.tif and dem_error/*bperp.tif are generated. It looks like the first three contain angles in units of radians (would degrees be a better choice for quick human interpretation?). Some thought needs to go in to the metadata to be included in the headers of these files; currently they're labelled as `DATA_TYPE=MULTILOOKED_IFG`, which isn't accurate. A DATA_UNITS entry could also be useful.
> 
> The function `pyrate.core.shared.read_lookup_table` is specific to a Gamma-format flat-binary FCOMPLEX lookup table file. I therefore think it would be better housed in `pyrate.core.gamma` and executed during `prepifg`.

Thanks Matt. I've worked on the tests today and the Travis tests should now (hopefully) be all passed. I've also fixed the DATA_TYPE for the three angle tifs and the bperp tifs. I'll move function `read_lookup_table` from `shared.py` to `gamma.py` in the next PR as discussed. The functions to calculate the various angles for the PyRate-multi-looked outputs (i.e. `get_lonlat_coords`, `get_radar_coords`, `calc_local_geometry` and `vincinv`) could be hosted in a new python file `geometry.py` which is called by the end of `prepifg.py`. Would that make sense to you?

mcgarth:
> Thanks Matt. I've worked on the tests today and the Travis tests should now (hopefully) be all passed. I've also fixed the DATA_TYPE for the three angle tifs and the bperp tifs. I'll move function `read_lookup_table` from `shared.py` to `gamma.py` in the next PR as discussed. The functions to calculate the various angles for the PyRate-multi-looked outputs (i.e. `get_lonlat_coords`, `get_radar_coords`, `calc_local_geometry` and `vincinv`) could be hosted in a new python file `geometry.py` which is called by the end of `prepifg.py`. Would that make sense to you?

Yes, that makes sense, to have these functions together in a module `geometry.py` (back in the day there was a `geodesy.py` module!). Also, it would be logical to have a test module `test_geometry.py` for unit testing these functions. You might also consider whether the functions `pyrate.core.refpixel.convert_pixel_value_to_geographic_coordinate` and `pyrate.core.refpixel.convert_geographic_coordinate_to_pixel_value` should be re-homed.

PR #303, opened 2020-10-15
Fix orbital corrections

basaks:
@mcgarth 23 July 2021: This PR implements significant changes to the orbital module as follows:

- new tests that construct synthetic orbital errors and check that the independent and network methods are both able to recover the input parameters.
- fixes to the network method and independent methods that were uncovered during the initial reviews of this PR (and confirmed with the new tests)
- introduce the `intercept` option in order to remove the previous dual use of the `offset` parameter (for the _intercept term_ in the DM and the _offset_ determination as the median of the ifg)
- introduce the DM `scale` as a user settable option for network method, default to 100 (a scale is not applied to the DMs in the independent method)
- parallelism is possible for independent methods
- multi-looking is possible for both independent and network methods

@basaks original description of the PR (titled **Add multiprocessing and multilooking capability to independent orbital correction**) as follows:

1. This PR add multilooking capability during independent orbital correction.
2. This PR also adds multiprocessing to independent orbital correction method.

It uses a similar least squares solution, except the `model` is based on only the interferogram's own multilooked phase data. The usefulness of this multilooking while using independent method should be compared against:

The independent method is trivially parallel and memory issues are not a concern.
Using multilooking during independent correction requies the following additional steps:

1. Create a in-memory (or on disc) multilooked file. We create the multilooked file in memory and assumed that it will be faster than creating it on disc.
2. Create an additonal deisgn matrix same size as that of the original ifg.
3. Too many nans in any one of the ifgs may make the process (least square solution) unstable.

Now the implementation is available in pyrate in case someone finds an use.


pyrate/core/orbital.py
```diff
@@ -221,23 +246,24 @@ def independent_orbital_correction(ifg, params):
         model = lstsq(clean_dm, data)[0]  # first arg is the model params
 
         # calculate forward model & morph back to 2D
+        original_dm = get_design_matrix(original_ifg, degree, offset)
```
basaks, line 249: Note this additonal design matrix requirment @mcgarth

mcgarth:
@mcgarth  and @chandra2ga will review this one

chandra2ga (commented):
Here are my observation and feedbacks while running independent orbital correction method and comparing the result with network based approach.

1) I can say that multi-looking capability and multi-processing steps for independent method are working. But major concern is while comparing two different methods I got some contradictory results. See figure 1 and 2. 
2) In fig-1 the phase ramp orientation shows opposite but not always the case and varying differently for other interferograms.
3) In fig-2 ranges are completely different with same colorbar and this I notices also for other interferograms of same dataset.

So we need further clarification and understanding which method correctly working and should select for orbital phase correction. 

Fig-1)
Fig1

Fig-2)
Fig2

tfuhrmann (changes requested):
I've run some tests to compare the orbital error estimates for different parameter settings and using both develop branch and sb/multiprocess branch. As pointed out by Chandra the results when using independent or network method are different (but an investigation whether these differences are acceptable is out of the scope of this PR). In addition to using the two different methods, I've also tested different orbit multi-look factors (1 and 10) as well as different orbfitdegrees (1, 2, 3). 

Observations:
- the results when using develop and sb/multiprocess are identical for an orbit ml-factor of 1.
- the comparison between ml-factor=1 and ml-factor=10 on the develop branch yields identical results for the independent and similar results for the network method.
- the results when using sb/multiprocess are identical to those when using develop branch for the network method, but different for the independent method (see statistics and plots below). This needs further investigation before the PR can be merged into develop.

**_Summary statistics_** using interferogram 20150703-20150727 in RSAT2 stack T081D (see /g/data/dg9/INSAR_ANALYSIS/CAMDEN/RSAT2/GAMMA/T081D/pyrate/out/)
Comparison of independent (1) and network method (2) of orbital correction using different polynomial degrees, orbit ml-factor: 1
_Develop branch:_
Orbital correction values min/max/mean: 
Method 1, degrees 1: -0.80/-0.21/-0.50
Method 1, degrees 2: 3.76/19.95/15.68
Method 1, degrees 3: -12.39/1.62/-1.36
Method 2, degrees 1: -0.80/-0.21/-0.50
Method 2, degrees 2: -9.43/1.50/-0.89
Method 2, degrees 3: -44.49/14.23/-5.92
_sb/multiprocess branch:_
Orbital correction values min/max/mean: 
Method 1, degrees 1: -0.80/-0.21/-0.50
Method 1, degrees 2: 3.76/19.95/15.68
Method 1, degrees 3: -12.39/1.62/-1.36
Method 2, degrees 1: -0.80/-0.21/-0.50
Method 2, degrees 2: -9.43/1.50/-0.89
Method 2, degrees 3: -44.49/14.23/-5.92
Comparison of independent (1) and network method (2) of orbital correction using different polynomial degrees, orbit ml-factor: 10
_Develop branch:_
Orbital correction values min/max/mean: 
Method 1, degrees 1: -0.80/-0.21/-0.50
Method 1, degrees 2: 3.76/19.95/15.68
Method 1, degrees 3: -12.39/1.62/-1.36
Method 2, degrees 1: -0.71/-0.18/-0.44
Method 2, degrees 2: -9.35/1.49/-0.84
Method 2, degrees 3: -43.41/13.33/-6.00
_sb/multiprocess branch:_
Orbital correction values min/max/mean: 
Method 1, degrees 1: -0.11/0.01/-0.05
Method 1, degrees 2: 1.07/4.59/3.64
Method 1, degrees 3: -2.79/0.40/-0.26
Method 2, degrees 1: -0.71/-0.18/-0.44
Method 2, degrees 2: -9.35/1.49/-0.84
Method 2, degrees 3: -43.41/13.33/-6.00

Comparison of orbit error estimates (left to right: orbiftdegrees 1, 2, 3) using the independent method resulting from develop branch (top line) and using sb/multiprocess branch (bottom line):
result_images_method1_develop_vs_sb-multiprocess

pyrate/core/orbital.py
```diff
@@ -81,82 +81,93 @@ def remove_orbital_error(ifgs: List, params: dict) -> None:
...
-                delayed(independent_orbital_correction)(ifg, params) for ifg in ifgs
+                delayed(independent_orbital_correction)(ifg_path, original_dm, params) for ifg_path in ifg_paths
             )
         else:
-            process_ifgs = mpiops.array_split(ifgs)
-            for ifg in process_ifgs:
-                independent_orbital_correction(ifg, params=params)
+            process_ifg_paths = mpiops.array_split(ifg_paths)
+            for ifg_path in process_ifg_paths:
+                independent_orbital_correction(ifg_path, design_matrix=original_dm, params=params)
```
mcgarth, line 120: Do we still need these two alternate parallelisation methods? I think we started to get rid of the use of `Parallel` calls in favour of the universal `array_split` method?

pyrate/core/orbital.py
```diff
@@ -81,82 +81,93 @@ def remove_orbital_error(ifgs: List, params: dict) -> None:
...
         # Here we do all the multilooking in one process, but in memory
         # can use multiple processes if we write data to disc during
         # remove_orbital_error step
         # A performance comparison should be made for saving multilooked
         # files on disc vs in memory single process multilooking
         if mpiops.rank == MAIN_PROCESS:
-            mlooked = __create_multilooked_dataset_for_network_correction(params)
+            mlooked = __create_multilooked_datasets(params)
             _validate_mlooked(mlooked, ifg_paths)
             network_orbital_correction(ifg_paths, params, mlooked)
```
mcgarth, line 131: multilooking the ifg datasets could be done in parallel (i.e. use array_split within `__create_multilooked_datasets`?)

mcgarth (changes requested):
The changes in this PR certainly improve the user experience and functionality of the orbital module. Some minor amendments are suggested.

Before merging we need to re-assess the impact of the changes (redo the experiments done by @chandra2ga and @tfuhrmann

pyrate/core/orbital.py
```diff
@@ -81,82 +81,93 @@ def remove_orbital_error(ifgs: List, params: dict) -> None:
...
+                 f'factors of X = {orbfitlksx} and Y = {orbfitlksy}')
 
     if method == INDEPENDENT_METHOD:
-        log.info('Calculating orbital correction using independent method')
-        #TODO: implement multi-looking for independent orbit method
-        if params[C.ORBITAL_FIT_LOOKS_X] > 1 or params[C.ORBITAL_FIT_LOOKS_Y] > 1:
-            log.warning('Multi-looking is not applied in independent orbit method')
-        ifgs = [shared.Ifg(p) for p in ifg_paths] if isinstance(ifgs[0], str) else ifgs
+        ifg0 = shared.Ifg(ifg_paths[0]) if isinstance(ifg_paths[0], str) else ifg_paths[0]
+        original_dm = get_design_matrix(ifg0, degree, offset)
```
mcgarth, line 111: I'm not sure about this. I think it is saying that the full-res DM from ifg0 (the first) will be used in the upscale of fullorb for every ifg? Could the ramifications of this be the cause of @tfuhrmann's observations about independent correction being different when using multi-looking factor > 1

pyrate/core/orbital.py
basaks: The gdal swig bindings prevent us from doing this. The multilooked ifgs are held in memory using in-memory tifs.

mcgarth: I have added this explanation as a comment in the code

pyrate/core/orbital.py
mcgarth: Disregard the above comment. The DM is the same for every ifg anyway.

richardt94:
It's not clear to me from the further commits/discussion since November last year whether @tfuhrmann's observations have been explained/resolved. Does the removal of the scale factor from the design matrix fix the issue with the multilooked and multiprocessed independent correction?

mcgarth:
> It's not clear to me from the further commits/discussion since November last year whether @tfuhrmann's observations have been explained/resolved. Does the removal of the scale factor from the design matrix fix the issue with the multilooked and multiprocessed independent correction?

To me, @tfuhrmann's observations show that there is a change to independent method results when using multi-looking >1. This is expected based on the scope of the PR (to enable multi-looking for the independent method). I have not fully assessed whether @tfuhrmann/@chandra2ga's observations are remedied by the latest changes to the PR. I think a full comparison/validation of independent and network methods is needed, but not in scope for approving this PR. Your thoughts @richardt94 ?

richardt94:
I think we shouldn't necessarily expect the methods to agree with each other, as they're using different input data, but we definitely need tests to confirm that they're working as expected. In particular - the orbital correction is fitting a linear, quadratic or cubic function to the data. Let's make some synthetic ifgs which are just linear/quadratic/cubic orbital errors plus some noise, and incorporate these in the test suite to confirm that all the orbit correction methods recover approximately correct orbital correction values.

tests/test_orbital.py
```diff
@@ -994,3 +985,39 @@ def orbfit_looks(request):
...
+        prepifg.main(params)
+        cls.params = Configuration(cls.conf).__dict__
+        correct._copy_mlooked(cls.params)
+        correct._create_ifg_dict(cls.params)
+
+    @classmethod
+    def teardown_class(cls):
+        shutil.rmtree(cls.params[C.OUT_DIR])
+
+    def test_independent_method_works_with_multilooking(self, orbfit_looks, orbfit_degrees, orbfit_method=1):
```
richardt94, line 1007: It looks like this test just checks that the correction actually executes and doesn't change the ifg shape. We should have some check of the value of the correction for a known synthetic ifg.

richardt94 (changes requested):
As I've outlined above, I think we need to start testing the corrections that we're getting out of the orbital fits with different methods. It's OK if they give different results for real interferograms with complicated combinations of signals, but we can test with a synthetic that's just "pure orbital error" and make sure that all of the methods (independent/network/multilook/single look) are effective at fitting this.

richardt94:
I have separated out the logic which calculates the networked orbital correction from the main network correction function, which has some undesirable file I/O stuff going on in it and really all we want to do is test that the logic successfully detects a function of the right form in a network of interferograms - we don't really care about checking that the interferograms can be saved afterwards because this is presumably elsewhere in the test suite.

This test passes for the example network I gave it (and I'm waiting on the full CI suite to check I didn't introduce any regressions). I will work on a similar test where I multilook the ifgs first and check that it still works, as the normal use case for the network orbital correction will be multilooked, and then revisit the independent method tests to see if they can be improved.

Before merging any of this, I'm a little worried that the design matrix appears to assign the constant coefficient to each _interferogram_ rather than each _epoch_ for the network inversion - is this intentional? It seems to me that the network method should be able to recover a model `orbit_error = a*x + b*y + c`, where the coefficients `a, b, c` are defined per-epoch.

pyrate/core/orbital.py
```diff
@@ -299,7 +299,7 @@ def __orb_inversion(design_matrix, data):
     B = design_matrix[~isnan(data)]
     d = data[~isnan(data)]
 
-    return dot(pinv(B, 1e-6), d)
+    return pinv(B) @ d
```
richardt94, line 302: Removing the artificial rcond restriction here allows us to solve the least-squares problem with more poorly-conditioned design matrices - this means we don't need to use a large "scale" factor in the tests any more.

richardt94:
I have implemented the network orbital correction using a constant parameter for each epoch rather than each interferogram, as discussed in my previous comment. This breaks some tests which compare against a reference implementation of the "old" method - I have skipped these. We should discuss whether this is the method we want to go with before merging the PR - in particular, the "old" and "new" network methods will interact differently with the reference pixel subtraction, which happens before any of the orbital corrections.

adeane-ga (commented):
Considering the changes that @richardt94 has made on this branch, I have rerun some PyRate comparisons to show latest output. 

I am hesitating to approve before we discuss the below comments. 

**Summary**

- the issue of a weak correction being applied due to multi-looking with the independent method has been resolved. 
- the network method is now producing the same result for all IFGs (caveat: I have not inspected all ~1450, but have looked through a similar amount as previous runs where I did pick up differences)
- Where we can see differences is when we plot the comparisons without subtracting the mean, but the only differences I have seen are an offset (i.e no orientation changes as seen previously)
- before merging, can someone explain this offset to me, I assume an offset does not matter due to the relative values across the 2D plane, but just want to confirm this. 
- we should also consider the last paragraph of this comment from @richardt94 : https://github.com/GeoscienceAustralia/PyRate/pull/303#issuecomment-877925853

**Details**

- frame is T16D_F31S_S1A
- full time series is processed
- orbit Fit is the only correction applied
- scale factor was set to 1 (instead of 100000)
- PDFs from which sample images are from are available on NAS: `T:\ops\insar\PROJECTS\PyRate\orbit_fit\ORBIT_FIT_PR`

**Processes are:**

1. Independent, multi-look=1
2. Independent, multi-look=5
3. Network, mult-look=5

**See example here of the three processes**

- Although I have not looked through every image ~1500, I cannot see any differences at all between the three methods.
- The issue of a low magnitude correction from the multi-looking in the independent method has been resolved.
- Do we expect there to be zero difference between methods? 

image

**Here we can see the same comparison but without the median subtracted**

- Note that the three results are still the same.

image

**He we can see differences in the offset in some IFGs**

- Note that this offset is not seen in all examples.
- Note that the resulting IFG is still the same as the other two processes. 

image

**This is the same IFGs but with the median subtracted**

image

adeane-ga (approved):
**Review**
- this review is approving the PR but with reference to future work.
- it addresses the goal of original PR issue outlined here: https://github.com/GeoscienceAustralia/PyRate/pull/303#pullrequestreview-515364579. 
- it includes information derived in addition to this comment: https://github.com/GeoscienceAustralia/PyRate/pull/303#pullrequestreview-708052099.

**Summary**

- For an inland frame (T16D_F31S-S1A), the independent method and the network method are producing consistent results (FIGURE 1). With the only difference being in the offset of the network correction – only seen when not subtracting mean from plots (FIGURE 2), is this expected? 
- For a frame containing more turbulent troposphere such as the coastal and topographically related atmospheric signals in (T147D_F37S_S1A), the network method still produces poor and inconsistent corrections in some interferograms (FIGURE 3) – expressed as overcorrections or corrections with a different orientation to the independent method. **This is not a blocker for merging this PR, but needs to be noted as the topic for for future work in a separate PR.**  
- The issue of a low magnitude correction when multi-looking in the independent method has been resolved. 
- Changing the scale factor when using the independent method has no affect – I believe this is by new design from @richardt94.
- Changing the scale factor in the network method does have an effect and is important. 
- Most of the corrections seem to be fitting to atmosphere, should we discuss running APS correction prior to orbit fit? 

**FIGURE (1)**
image

**FIGURE (2)**
image

**FIGURE (3)**
image

mcgarth (approved):
Based on my own code review, test runs, and @adeane-ga's detailed review, I believe we have fixed the issues in the Orbital module that were uncovered by the initial reviews by @chandra2ga and @tfuhrmann. To me, the correction looks sensible for both the independent and network methods, whereas previously it did not. I am therefore approving this PR for merging.

Some follow up considerations:
- communicate to users when is it worth using the orbital correction (what sensor, size of imaged area etc). I believe some of @adeane-ga's observations are because the Sentinel-1 data used does not contain appreciable "orbital error", so yes we are fitting over signals, such as the atmosphere.
- some more testing of the use of scale factor in the DM for network method. As @adeane-ga points out, changing the scale does impact the correction magnitudes massively.
- refactor the flow of the module and re-use functionality for both independent and network methods where possible (e.g. multi-looking)
- reconsider the need for the application of the `offset` as the median of the image during the orbital module (`pyrate.core.ref_phs_est` can be ran before and/or afterwards to have the same impact on the data)

PR #304, opened 2020-10-23
Tf/dem error

tfuhrmann:
Implemented DEM error estimation generating correction values for each interferogram in subfolder /out/dem_error as well as a /out/dem_error.tif file giving the average height difference w.r.t. the input DEM.

ToDo:
- further testing of the validity of resulting correction values
- implement tests for the new functionality (i.e. dem_error.py, geometry.py and read_lookup_table in gamma.py) once the new unit test data set has been established


pyrate/core/dem_error.py
```diff
@@ -14,299 +14,285 @@
...
+                np.save(file=os.path.join(params[cf.TMPDIR], 'dem_error_correction_{}.npy'.format(t.index)), \
+                        arr=tmp_array)
+
+            # wait for all processes to finish
+            mpiops.comm.barrier()
+
+            # write dem error and correction values to file
+            mpiops.run_once(_write_dem_errors, ifg_paths, params, preread_ifgs, tiles)
+            shared.save_numpy_phase(ifg_paths, params)
+
```
basaks, line 144: It's possbile to use the `tiles_split` function here. In order to do that you may need to look into tiled or IfgPart equivalent of the `Geometry` class.

tests/test_dem_error.py
```diff
@@ -0,0 +1,74 @@
...
+pyrate_bperp = np.array([33.48592183, 3.44669685, -75.37369399, -26.88597679, -33.25298942, -108.84360354, 3.74075472,
+                         -3.19700977, -14.58390611, 10.06920291, -51.12649599, -5.74544068, -17.36872483, -30.4772929,
+                         7.12691256, -37.68943916, -73.14248882, -11.45674522, -24.64851804, 12.69928323, -32.16248418,
+                         -20.86746046, 61.514626, 48.30928659, -13.17640207, 24.28126177, -36.84111057, -20.5870326,
+                         77.8291117, -8.66115426])
+
+
+# compare the GAMMA and PyRate Bperp estimates
+def test_pyrate_bperp_matches_gamma_bperp(gamma_bperp):
+    np.testing.assert_array_almost_equal(pyrate_bperp / 1e6, gamma_bperp / 1e6)  # max difference < 10mm
```
basaks, line 69: @tfuhrmann, Is this something you had in mind?
From here, we just need to make sure that `pyrate_bperp` is coming from pyrate code and not as in this array `pyrate_bperp` above.

tfuhrmann:
> Yes I agree that we need to roll back to Thomas' version of the `remove_file_if_exists` function. Also, I wonder if `remove_file_if_exists` function is better housed in shared.py? it's generic and used in two places (dem_error.py and geometry.py)

I've now changed `remove_file_if_exists` back to the old version and moved it into shared.py

tfuhrmann:
I think the Python 3.6 Travis checks fail because of these lines of code:
```
    # get azimuth angle value of PyRate file azimuth_angle.tif
    out = run(f"gdallocationinfo {tif_file} {x0} {y0}", shell=True, capture_output=True,
              text=True).stdout
    angle_value = float(out.split('Value:')[1].strip('\n'))
```
Looks like `capture_output=True` is only for Python version 3.7+ (see https://stackoverflow.com/questions/52219836/subprocess-run-doesnt-return-stdout-or-stderr)

pyrate/core/geometry.py
```diff
@@ -92,9 +108,17 @@ def calc_radar_coords(ifg, params, xmin, xmax, ymin, ymax):
...
+def get_sat_positions(lat: np.ndarray, lon: np.ndarray, epsilon: np.ndarray,
+                      heading: np.float64, roll: np.float64) -> Tuple[np.ndarray, np.ndarray]:
     """
-        Function to calculate the lon/lat position of the satellite for a given pixel location
+    Function to calculate the lon/lat position of the satellite for each pixel.
+    :param lat: Ground latitude for each pixel (decimal degrees).
+    :param lon: Ground point longitude for each pixel (decimal degrees).
+    :param epsilon:
+    :param heading: Satellite flight heading (radians).
+    :param roll: Right or left look direction w.r.t. satellite heading; +ve = right looking (radians).
```
tfuhrmann, line 119: I don't think 'roll' is a good choice for this parameter name. The roll angle in aviation describes a motion about the longitudinal axis (i.e. the flight direction axis), which is not what we have here. What about 'look_direction' or 'look_dir'?

See also ESA Glossary here https://earth.esa.int/handbooks/asar/CNTR5-2.html on the SAR look direction:
The radar look direction defines the angle in the horizontal plane in which the radar antenna is pointing when transmitting a pulse and receiving the return signal from the ground or from an object. The look direction is an angular measurement (in degrees) and is usually made with respect to true North. In side-looking imaging radar (SLAR), the look direction is often orthogonal (normal) to the flight trajectory (azimuth) of the platform carrying the radar and is thus synonymous with the range direction. The radar look direction is an important parameter when analysing features with a preferred orientation, for example fracture patterns in rock formations, regular street patterns, or ocean waves, as these may be enhanced through choice of appropriate radar illumination direction. ( See also "Imaging Geometry" in the geometry glossary ).

pyrate/core/geometry.py
```diff
@@ -128,19 +162,21 @@ def calc_pixel_geometry(ifg, rg, lon, lat, params):
...
     DEM_data.open(readonly=True)
     dem_height = DEM_data.height_data
 
-    # convert angles to radians
+    # convert to radians
     lon = np.radians(lon)
     lat = np.radians(lat)
     heading = np.radians(heading)
-    azimuth = np.radians(azimuth)
+    right_or_left_look = np.radians(right_or_left_look)
```
tfuhrmann, line 180: maybe replace with 'look_direction'

pyrate/core/geometry.py
mcgarth: I think your suggestion of `look_dir` is much better than my `roll`!

pyrate/core/geometry.py
mcgarth: I prefer `look_dir`, as in above comment

mcgarth (approved):
I have done some thorough testing of this branch over a period of about a ~week, including as it evolves over that period. I have committed minor small changes to improve the "user interface" with the new functionality. But my opinion is that the core new functionality to estimate DEM errors (including geometry and perpendicular baseline calculations) is well put together and really well documented. It is a great addition to the PyRate software - well done @tfuhrmann !

PR #311, opened 2020-12-22
Fix spatial temporal filter

mcgarth:
This PR fixes the spatio-temporal filter functionality (that previously produced spurious results) in `pyrare.core.aps` and puts the functionality under unit test.
- change temporal filter cutoff to units of days (rather than years); default 30 days
- remove filter types and only retain only the **gaussian** option for spatial and temporal filter
- remove superfluous config options slpfmethod, slpforder, tlpfmethod
- convert spatial cutoff distance (default 1 km) to a scaled "sigma" value for use in gaussian kernel
- spatial interpolation by nearest neighbour is on by default; needed to reduce edge effects in FFT used by spatial filter
- Implement unit tests for temporal and spatial gaussian filter options. Compare bespoke code against `scipy.ndimage` equivalents (these functions only support regularly sampled data in time and space, whereas the functions implemented in `pyrate.core.aps` support irregular temporal sampling and different x and y spatial resolutions)


basaks (approved):

chandra2ga:
I reviewed this PR and performed different unit tests. Now I can confirm that this PR is working as expected and ready to merge.

Here are some feedback:
1. 1- 4 comments are checked. 
2. 5th comment regarding spatial interpolation, it's good to have nearest neighbor method as default option but I would definitely suggest to keep other options open for users to choose, and might suitable for their processing.
3. I checked and performed unit test for temporal and spatial gaussian filters option. Looks alright to me.

So I approved this PR and ready to merge into develop/ 👍

chandra2ga (approved):

mcgarth:
To be clear, `nearest` is the default option for spatial interpolation in the APS filter, but `cubic` and `linear` methods are still selectable via the `slpnanfill_method` config parameter option.
I have removed `slpnanfill_method` from the top level `input_parameters.conf` because I believe `nearest` to be the best choice. But those users who are in the know, can add that parameter into their config file and choose `linear` or `cubic` options instead. ;-)

PR #314, opened 2021-01-20
Add [skip ci]

jmettes:
It works: https://github.com/GeoscienceAustralia/PyRate/actions/runs/497579479

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,6 +6,7 @@ jobs:
   build:
     runs-on: ubuntu-16.04
     name: Python ${{ matrix.python }}
+    if: "!(contains(github.event.head_commit.message, '[skip ci]') || contains(github.event.pull_request.head_commit.message, '[ci skip]'))"
     strategy:
       matrix:
         include:


jmettes:
Hmm still burning some carbon in the "Checks" above

jmettes:
It looks the CI will still always run on pull requests, even if "skip ci" is included inside the commit message.

This is because the approach for skipping relies on checking the commit messages inside the [event metadata](https://docs.github.com/en/developers/webhooks-and-events/webhook-events-and-payloads#pull_request), which triggered the CI. It turns out that pull request event metadata does not contain a list of commit messages, so there's no way to check if they contain "skip ci".

We could check for the title of the PR, and include "skip ci" in there.

Alternatively, a "pre_ci" workflow step could be added to manually parse git commit messages, and then used to skip subsequent workflow steps:
https://github.community/t/accessing-commit-message-in-pull-request-event/17158/9

basaks:
@jmettes , thank you for confirming the change works. I will merge it in now.

I will checkout the `pre_ci` step you tagged in the commit above. It looks promising.

PR #316, opened 2021-02-10
make mpi an optional dependency

basaks:
1. adds a non-mpi system support
2. drops support for python3.6 env
3. new ci env to track non-mpi working env


mcgarth (commented):
Working with fc57af5

I could get PyRate to work with/without MPI on gadi head-node with the following instructions (slightly modified from here: https://geoscienceaustralia.github.io/PyRate/installation.html#hpc), but only if I removed `mpi4py==3.0.3` line from `requirements.txt`

```
module purge
module load python3/3.7.4
module load gdal/3.0.2
export PYTHONPATH=/apps/gdal/3.0.2/lib64/python3.7/site-packages:$PYTHONPATH
python3 -m venv ~/TestVenv
source ~/TestVenv/bin/activate
pip install -r requirements.txt
python3 setup.py install
pyrate workflow -f input_parameters.conf

module load openmpi/4.0.2
pip install mpi4py==3.0.3
python3 setup.py install
mpirun -n 4 pyrate workflow -f input_parameters.conf
```

When doing `pip install -r requirements.txt` with `mpi4py` **in** the requirements, the pip install would fail because no MPI in the environment and therefore bindings could not be built. The simple fix was to pull `mpi4py` from the requirements, and do a manual `pip install mpi4py==3.0.3` when needed (after `module load openmpi`), prior to the `python3 setup.py install` step

mcgarth:
Working with f3882aa1 

The above commits get around the issue I mentioned above.

New issue: `requirements.txt` is overwritten as part of `python setup.py install`. I don't think this is ideal, is there another way?

I approve the merge, but with the caveat that we address the overwriting of `requirements.txt` in the near future

PR #318, opened 2021-02-21
remove config.py and test for 6 and 8 digit filenames [ci skip]

basaks:
1. remove config.py, associated tests and refactor required functions into `configuration`, `shared`, and constants into `constants`.
2. tests for 6 and 8 digits filenames (in new test file test_constants.py)
3. Mexico Crop A data now under `test_coherence`.


tests/test_pyrate.py
```diff
@@ -323,12 +328,12 @@ def test_stack_rate_equal(self):
 
 
 @pytest.mark.slow
-@pytest.mark.skipif(not PYTHON3P8, reason="Only run in python 3.8")
+@pytest.mark.skipif(not PY37GDAL304, reason="Only run in python 3.8")
```
mwheeler, line 331: The skip reason doesn't match what's being done anymore... is this intended?

tests/test_mpi_vs_multiprocess_vs_single_process.py
```diff
@@ -23,16 +23,17 @@
...
 
 from tests.common import (
     assert_same_files_produced,
     assert_two_dirs_equal,
     manipulate_test_conf,
     MEXICO_CROPA_CONF,
     PY37GDAL304,
-    PY38GDAL304
+    PY38GDAL304,
+    PYTHON3P7
```
mwheeler, line 36: `PYTHON3P7` has been added to import list, but unused... is this intentional / did some change get missed that was supposed to use this?

mwheeler:
This was a mammoth PR (guessing it was mostly done via sed or some other search/replace, as that's what most of the edits where) - as a result I only reviewed the exact changes made (I didn't dig too much into related code).

No major issues found, just the two comments above.

mwheeler (approved):

tests/test_pyrate.py
basaks: hey, thank you for the review. Yes, it's an oversight from me. We just redistributed the long running tests into the various environements, but I will change the text!

tests/test_mpi_vs_multiprocess_vs_single_process.py
basaks: same effort, probably changed my mind last minute to run the test in a different env.

mcgarth (approved):
I have viewed (at a high-level) the changes and they look good to me. Due to the nature of the PR (a refactor), I am basing my approval mostly on the passing of the test suite, and its good! Thanks @basaks

PR #319, opened 2021-03-04
Add switch to specify uncertainty communication in rate/velocity error maps

richardt94:
1. I have added a switch to the config file, which allows the user to specify whether they would like the velocity error maps to represent 1- or 2-sigma errors. This switch applies to the error maps produced by both `stack` and `timeseries` methods. The default is 2-sigma (95% confidence interval) because this is the standard for uncertainty communication in many fields of science.
2. I modified the test suite, specifically `test_timeseries.py` and `test_stackrate.py`, to manually specify 1-sigma errors in the default parameters dictionary. This is for compatibility with the legacy test datasets which have 1-sigma error maps.
3. I added a new test to `test_timeseries.py` which confirms that the switch actually produces a 2-sigma error map when this is specified.

### DIFF
--- a/input_parameters.conf
+++ b/input_parameters.conf
@@ -22,6 +22,12 @@ savenpy:       0
 savetsincr:    0
 
 #%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+# Integer parameters
+
+# Number of sigma to report velocity error. 1 or 2. Default: 2 (TIMESERIES/MERGE)
+velerror_nsig: 2
+
+#%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 # Multi-threading parameters used by correct/stacking/timeseries
 # gamma prepifg runs in parallel on a single machine if parallel = 1
 # parallel: 1 = parallel, 0 = serial
--- a/input_parameters.conf
+++ b/input_parameters.conf
@@ -24,7 +24,7 @@ savetsincr:    0
 #%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 # Integer parameters
 
-# Number of sigma to report velocity error. 1 or 2. Default: 2 (TIMESERIES/STACK)
+# Number of sigma to report velocity error. Positive integer. Default: 2 (TIMESERIES/STACK)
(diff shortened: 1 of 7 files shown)


tests/test_stackrate.py
```diff
@@ -199,7 +200,7 @@ def test_stack_rate(self):
 
     def test_stackrate_error(self):
         """
-        Compare with legacy data
+        Compare with legacy data. Default behaviour is now 2-sigma, so mult legacy by 2.
```
mcgarth, line 203: It's not clear to me exactly where "mult legacy by 2" is actually happening in the test code. Is this comment line still relevant?

input_parameters.conf
```diff
@@ -21,6 +21,12 @@ savenpy:       0
 # Optional save of incremental time series products (TIMESERIES/MERGE)
 savetsincr:    0
 
+#%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+# Integer parameters
+
+# Number of sigma to report velocity error. 1 or 2. Default: 2 (TIMESERIES/MERGE)
```
mcgarth, line 27: I think this parameter is only relevant to TIMESERIES step, since it is applied in `pyrate/core/timeseries.py`. Delete "MERGE" in this comment

mcgarth:
Thanks @richardt94 , this is a good change to the code! I have a couple of suggested additions to your functionality. Depending how much effort they require you can include in this PR, or save for a later ticket.
- do we need to constrain n-sigma to only 1 or 2? What about any integer >1 ? (Default of 2 is sensible though!)
- you could add a geotiff tag to the two error products `stack_error.tif` and `linear_error.tif`, giving the value of n-sigma

mcgarth (approved):

richardt94:
I have rebased the changes on top of the latest `develop` and removed the restriction to only one or two sigma - can now be an arbitrary positive integer. I have not added metadata to the output geotiff because I can't understand the system for getting metadata into the geotiffs - it looks like `merge` is opening some stored metadata that is generated by `correct` and then only re-writing specific tags that are defined in `ifgconstants.py`? This represents a good opportunity for code rationalisation IMO, possibly as part of PIS-169 (merging products in memory instead of writing/reading to disk as an intermediate step).

@basaks this PR is now ready for your review, then if you approve I will squash and merge.

basaks (approved):

mcgarth:
velocity error nsigma metadata has been added to the output error geotiffs in https://github.com/GeoscienceAustralia/PyRate/pull/327

PR #320, opened 2021-03-08
Sb/calc coherence image stats [ci skip]

basaks:
1. write coherence statistics files
2. tests to check coherence files are created in mpi/multiprocess/single process jobs and compare them


pyrate/prepifg.py
```diff
@@ -78,22 +79,42 @@ def main(params):
...
+    if params[C.LT_FILE] is not None:
         log.info("Calculating and writing geometry files")
-        mpiops.run_once(_write_geometry_files, params, exts, transform, ifg_paths[0])
+        mpiops.run_once(__write_geometry_files, params, exts, transform, ifg_paths[0].sampled_path)
     else:
         log.info("Skipping geometry calculations: Lookup table not provided")
 
+    if params[C.COH_FILE_LIST] is not None:
+        log.info("Calculating and writing geometry files")
+        mpiops.run_once(__calc_coherence_stats, params, ifg_paths[0].sampled_path)
```
mcgarth, line 90: Can this be extended to run with `tiles_split`? It'll be a big computation to hold in memory when we need to hold hundreds of coherence files in memory, each with 1M pixels.

mcgarth:
I ran the new coherence statistics functionality over the Mexico CropA data - the outputs seem to make sense:
image

NMoghaddam (approved):
I reviewed the newly developed functionality in PyRate to calculate the mean, median and standard deviation of coherence files in a stack of interferograms. 
The functionality works properly for the Mexico test dataset. In addition to check the output in QGIS, I also did a quantitative assessment with the previously developed python code in the jupyter lab and compare the two as demonstrated here. I can confirm that the compatibility is very close and accepted. 
[Coherence_Statistics.docx](https://github.com/GeoscienceAustralia/PyRate/files/6105778/Coherence_Statistics.docx)
Accordingly, I approve this review request. Thank you Sudipta.

mcgarth (approved):
Thanks @basaks for this new functionality, which I think produces sensible output. Thanks also for addressing my review comments.
An outstanding improvement to be implemented would be doing the coherence stats calculations in tiles, to exploit parallellism of the calculation (done on a pixel-by-pixel basis). This will be needed for large InSAR datasets that we expect to be processing shortly

pyrate/prepifg.py
mcgarth: To be implemented in a separate PR

PR #321, opened 2021-03-11
Add LOS projection options for supported outputs

basaks:
This PR contains:
1. LOS projection options (los, pseudo_vertical, pseudo_horizontal) for {'tsincr', 'tscuml', 'linear_rate', 'stack_rate'} outputs.
2. los projection is controlled by new param option `los_projection`
3. New metadata introduced to identify which los projection option was used:  `LINE-OF-SIGHT`, `PSEUDO-VERTICAL`, `PSEUDO-HORIZONTAL`.
3. moved geometry files into `geometry_dir`.
4. remove some inconsistencies around definitions/use of `orb_error_dir`, `dem_error_dir` and `aps_error_dir`.
5. Remove tests from `test_coherence.py` as we don't have a `coherence.py`, and moved these tests to more appropraite locations (test_gdal_python.py) following `pyrate` convention.
6. Geometry output file name denifinitions are now centralised in `Configuration` class.
7. Some further code rationalisation and optimisation.


pyrate/merge.py
```diff
@@ -248,21 +245,43 @@ def assemble_tiles(s: Tuple, dir: str, tiles: Tile, out_type: str, index: Option
...
+    'linear_rsquared': ifc.LINRSQ,
+    'tsincr': ifc.INCR,
+    'tscuml': ifc.CUML
+}
+
+los_projection_out_types = {'tsincr', 'tscuml', 'linear_rate', 'stack_rate'}
+los_projection_divisors = {
+    ifc.LINE_OF_SIGHT: lambda data: 1,
+    ifc.PSEUDO_VERTICAL: lambda data: np.sin(data),
+    ifc.PSEUDO_HORIZONTAL: lambda data: np.cos(data)
```
mcgarth, line 276: Following some thought on this (including hand drawn sketches!), I believe we have these wrong. The incidence angle is measured between the LOS vector and the nadir (vertical) direction at the ground location. So based on this we should have:
```
los_projection_multipliers = {
    ifc.LINE_OF_SIGHT: lambda data: 1,
    ifc.PSEUDO_VERTICAL: lambda data: np.cos(data),
    ifc.PSEUDO_HORIZONTAL: lambda data: np.sin(data)
```
The result of these multiplications with the input signal is a projected signal that is *smaller* than the LOS signal (LOS on the hypotenuse of right-angled triangle)

pyrate/merge.py
```diff
@@ -272,44 +291,37 @@ def _save_merged_files(ifgs_dict, outdir, array, out_type, index=None, savenpy=N
...
+    if out_type in los_projection_out_types:  # apply LOS projection for these outputs
+        incidence_path = Path(Configuration.geometry_files(params)['incidence_angle'])
+        if incidence_path.exists():  # We can do LOS projection
+            if params[C.LOS_PROJECTION] == ifc.LINE_OF_SIGHT:
+                log.info(f"Retaining Line-of-sight signal projection for file {dest}")
+            elif params[C.LOS_PROJECTION] != ifc.LINE_OF_SIGHT:
+                log.info(f"Projecting Line-of-sight signal into {ifc.LOS_PROJECTION_OPTION[params[C.LOS_PROJECTION]]} for file {dest}")
+            incidence = shared.Geometry(incidence_path)
+            incidence.open()
+            array /= los_projection_divisors[params[C.LOS_PROJECTION]](incidence.data)
```
mcgarth, line 312: From the other comment, I think this needs to be:
`array *= los_projection_multipliers[params[C.LOS_PROJECTION]](incidence.data)`

mcgarth (changes requested):
Generally the functionality is working as we envisaged, except I think pseudo-horizontal and pseudo-vertical signals need to be *smaller* (not larger) than the LOS signal - see more detailed code-level comments.

I'd also like to see a unit test in `test_merge.py` that checks the mathematical operations have been applied correctly - i.e. for a given incidence angle of say 30 deg, test that LOS > Vert > Horz, or even test precise values equal those expected for a 1 pixel toy problem.

pyrate/merge.py
mcgarth: After some discussion, I have realised that my suggestion is incorrect, and the implemented code is correct. So confusing! After projection to vertical, the signal will be *larger* than the LOS signal, because the vertical forms the hypotenuse side of the projection triangle.

pyrate/merge.py
mcgarth: My comment is incorrect. The code is implemented correctly

richardt94 (changes requested):
I agree with Matt's comment that `sin` and `cos` are in the wrong place for vertical and horizontal. Specific comments are on `merge.py`.

richardt94 (approved):

pyrate/merge.py
mcgarth: That's a good way to explain the scenario. Thanks @richardt94 !

Regarding "pseudo-horizontal", you are correct. The motion is in the horizontal plane (perpendicular to nadir). A future extension could be to actually estimate the pseudo-E-W or pseudo-N-S motion by bringing the azimuth angle in to the calculation. Since I don't think we will use this, I think it is low priority - maybe a suggestion for an external contributor on Github Issues page? https://github.com/GeoscienceAustralia/PyRate/issues

mcgarth (approved):

mcgarth:
Once the CI is passing, this PR can be merged

PR #327, opened 2021-03-25
signal polarity configuration implemented and tested

basaks:
@mcgarth edit:
This PR implements a simple signal polarity conversion, which is applied to output geotiff products generated during the MERGE step. Data sign is not affected in the other processing steps. Default is multiplying the output data arrays by -1. This ensures GA's standard input interferograms are converted such that PyRate outputs have range increases as negative and range decreases as positive.
- Metadata for `signal_polarity` and `velerror_nsig` has been added to the geotiffs
- Info messages have been added to MERGE step to report on conversions being applied

### DIFF
--- a/input_parameters.conf
+++ b/input_parameters.conf
@@ -1,7 +1,7 @@
 # PyRate configuration file for GAMMA-format interferograms
 #
 #%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
-# Optional ON/OFF switches - ON = 1; OFF = 0
+# Optional Correction ON/OFF switches - ON = 1; OFF = 0
 
 # Coherence masking (PREPIFG)
 cohmask:       0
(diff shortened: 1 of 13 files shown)


input_parameters.conf
```diff
@@ -103,6 +103,13 @@ noDataValue:   0.0
 # Nan conversion flag. Set to 1 if missing No-data values are to be converted to NaN
 nan_conversion: 1
 
+
+# sign convention for phase data
```
richardt94, line 107: Correct comments - subsidence as negative and uplift as positive mean the same thing.

Also maybe add a disclaimer that the sign convention only works as stated if inputs are in standard GAMMA format?

tests/test_merge.py
```diff
@@ -96,54 +102,74 @@ def test_los_conversion_comparison(self):
...
             k_dir = Path(params[C.OUT_DIR]).joinpath(ifc.LOS_PROJECTION_OPTION[k])
             k_dir.mkdir(exist_ok=True)
-
-            for out_type in los_projection_out_types:
-                for tif in Path(params[C.OUT_DIR]).glob(out_type + '*.tif'):
-                    shutil.move(tif, k_dir.joinpath(tif.name))
             all_dirs[k] = k_dir
+            self.run_with_new_params(k_dir, params)
+
+        signal_dir = Path(params[C.OUT_DIR]).joinpath('signal_polatiry_dir')
```
richardt94, line 113: Typo, polatiry -> polarity

pyrate/merge.py
```diff
@@ -310,7 +310,7 @@ def __save_merged_files(ifgs_dict, params, array, out_type, index=None, savenpy=
                          f"{ifc.LOS_PROJECTION_OPTION[params[C.LOS_PROJECTION]]} for file {dest}")
             incidence = shared.Geometry(incidence_path)
             incidence.open()
-            array /= los_projection_divisors[params[C.LOS_PROJECTION]](incidence.data)
+            array /= los_projection_divisors[params[C.LOS_PROJECTION]](incidence.data) * params[C.SIGNAL_POLARITY]
```
richardt94, line 313: The sign convention is applied in merge, but in the example configs it is listed as a parameter for prepifg

richardt94 (approved):
Looks good to me. Some changes/explanation needed in the config file comments as noted.

input_parameters.conf
mcgarth: N.B. there is no "standard" GAMMA format ;-)

richardt94: Well in that case maybe we want something like this?
```
+1 - keep sign convention of input interferograms
-1 - reverse sign convention compared to input interferograms
```
We don't want a user putting in interferograms where negative values already correspond to motion away from the satellite, and expecting `signal_polarity: -1` to preserve that.

mcgarth: Yes great point. We still need our user to understand what their input data signal convention is.

pyrate/merge.py
mcgarth: I have removed the extra comments from backend test configs, and updated the top level config to show it is applied in MERGE

input_parameters.conf
mcgarth: I have updated `input_parameters.conf` based on @richardt94's idea

mcgarth (approved):
I made some additions. @richardt94 to re-review please

richardt94 (approved):
All looks good to me. Velocity error sigma values are now included in the geotiff metadata which is a nice addition too.

PR #329, opened 2021-04-29
Add support for GAMMA initial baseline estimates

mcgarth:
This PR adds support for reading and using the Initial baseline estimates from GAMMA `*base.par` files.

The Precise baseline estimate will be used if it is found (i.e. if one of the TCN components is non-zero), otherwise the Initial baseline estimates will be read.

This change is required to support interferograms where precise baseline estimation has been skipped (i.e. when the orbital state vectors are deemed adequate to solve for the orbital phase contribution. This is usually the case for Sentinel-1).

### DIFF
--- a/pyrate/core/gamma.py
+++ b/pyrate/core/gamma.py
@@ -233,42 +233,39 @@ def parse_baseline_header(path: str) -> dict:
     Will read the Precise baseline estimate, if available,
     otherwise will read the Initial baseline estimate.
 
-    :param path: `Full path to Gamma *base.par file`
+    :param path: Full path to Gamma *base.par file
 
     :return: bdict: Dictionary of baseline values
     """
-    lookup = _parse_header(path)
+    lookup = _parse_header(path)  # read file contents in to a dict
+
+    # split the initial and precise baselines
+    initial = lookup[GAMMA_INITIAL_BASELINE]
+    initial_rate = lookup[GAMMA_INITIAL_BASELINE_RATE]
+    precise = lookup[GAMMA_PRECISION_BASELINE]
+    precise_rate = lookup[GAMMA_PRECISION_BASELINE_RATE]
+
+    # read the precise baseline if all components are non-zero
+    # otherwise read the initial baseline
+    if float(precise[0]) == 0.0 and float(precise[1]) == 0.0 and float(precise[2]) == 0.0:
+        log.debug('Reading Initial GAMMA baseline values')
+        baseline, baseline_rate = initial, initial_rate
+    else:
+        log.debug('Reading Precise GAMMA baseline values')
+        baseline, baseline_rate = precise, precise_rate
+
+    # Extract and return a dict of baseline values
+    bdict = {}
 
     # baseline vector (along Track, aCross track, Normal to the track)
-    initial_tcn = lookup[GAMMA_INITIAL_BASELINE]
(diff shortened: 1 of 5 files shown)


pyrate/core/gamma.py
```diff
@@ -226,29 +227,45 @@ def parse_dem_header(path):
...
+
+    # split the initial and precise baselines
+    initial = lookup[GAMMA_INITIAL_BASELINE]
+    initial_rate = lookup[GAMMA_INITIAL_BASELINE_RATE]
+    precise = lookup[GAMMA_PRECISION_BASELINE]
+    precise_rate = lookup[GAMMA_PRECISION_BASELINE_RATE]
+
+    # read the precise baseline if all components are non-zero
+    # otherwise read the initial baseline
+    if float(precise[0]) == 0.0 and float(precise[1]) == 0.0 and float(precise[2]) == 0.0:
```
richardt94, line 250: Is there any situation where one of the baseline coordinates might be precisely zero, or does this always indicate that we should use the initial estimate instead?

Comment says _all_ components non-zero, code tests if _any_ components are non-zero.

tests/test_gamma.py
```diff
@@ -363,3 +363,54 @@ def test_meta_data_exists(series_ifgs, parallel_ifgs):
...
+                 'BASELINE_RATE_N': -0.0055098}
+
+        exp_p = {'BASELINE_T': 0.0, 'BASELINE_C': -103.8364725,
+                 'BASELINE_N': 2.8055662, 'BASELINE_RATE_T': 0.0,
+                 'BASELINE_RATE_C': -0.0182215,
+                 'BASELINE_RATE_N': -0.0065402}
+
+        # Precise values are read
+        assert self.prec != exp_i
+        assert self.prec == exp_p
```
richardt94, line 391: Should this test pass? The comment in `parse_baseline_header` suggests that precise values are ignored unless _all_ values are non-zero.

richardt94 (changes requested):
Looks good, but the code and test suite for checking when to use precise baseline doesn't match the logic in the comments. Please choose one and update code or comment to match.

richardt94 (approved):
Looks good - code and comments now match.

tests/test_gamma.py
```diff
@@ -363,3 +363,54 @@ def test_meta_data_exists(series_ifgs, parallel_ifgs):
...
+                 'BASELINE_N': 0.0, 'BASELINE_RATE_T': 0.0,
+                 'BASELINE_RATE_C': 0.0, 'BASELINE_RATE_N': 0.0}
+
+        # Precise values are ignored
+        assert self.prec != exp_i
+        assert self.prec != exp_p
+
+        # Initial values are read
+        assert self.init == exp_i
+        assert self.init != exp_p
```
basaks, line 415: for an equality check in a dict, only one element mismatch will result true. Do you need all  elements to be different? Or  is just one unequal element in the dict is enough here?

mcgarth: Thanks for questioning @basaks. Because we are testing against hard-coded truths here (`exp_i` and `exp_p` values are derived from the files that are being read in the tests: `20160114-20160126_base_init.par` and `20160114-20160126_base.par`), it's either all values (i.e. 6) equal or no values equal, nothing in between. So I think it should be ok?

basaks (approved):

PR #330, opened 2021-05-03
Deprecate OBS_DIR, SLC_DIR and COH_FILE_DIR parameters

basaks:
OBS_DIR, SLC_DIR and COH_FILE_DIR are no longer used as config parameters. These params are removed from code and config files.

WORKING_DIR replaces OBS_DIR in the test suite


richardt94 (commented):
This successfully removes the stated parameters from the code, with the exception of `OBS_DIR` which is renamed to `WORKING_DIR`. All looks good to me, but if `WORKING_DIR` is still a valid part of the configuration, should it be exposed/added to the example `.conf` files?

basaks:
> This successfully removes the stated parameters from the code, with the exception of `OBS_DIR` which is renamed to `WORKING_DIR`. All looks good to me, but if `WORKING_DIR` is still a valid part of the configuration, should it be exposed/added to the example `.conf` files?

Hey, working dir is not a valid part of the conf file. You may choose to 

> This successfully removes the stated parameters from the code, with the exception of `OBS_DIR` which is renamed to `WORKING_DIR`. All looks good to me, but if `WORKING_DIR` is still a valid part of the configuration, should it be exposed/added to the example `.conf` files?

@richardt94 the string working_dir is only used in the tests, and as such I have moved it into the `tests.common` file. 
We don't need to expose it to the user. May be this is more suitable as not part of the core pyrate files.

richardt94:
Definitely happier with having the `WORKING_DIR` parameter in tests rather than the core files - I approve the changes now

mcgarth:
closing without merging to `develop`, since this will occur when https://github.com/GeoscienceAustralia/PyRate/pull/331 is merged

PR #331, opened 2021-05-05
Restructure outdir

basaks:
This PR redirects the timeseries and stack outputs from the busy output directory to the new timeseries_dir and velocity_dirs.

Redirected outs include all associated tifs,  numpy outputs, and other plots like the  .pngs, and the .kmls.


richardt94 (approved):
Changes appear correct to me. I note that we're using both `pathlib` and `os.path` in `merge.py` (and possibly elsewhere) - probably a legacy of older code, we may want to switch to only using `pathlib` as a code rationalisation step in future.

mcgarth (approved):
Thanks @basaks, the changes have the desired effect

Issue #310, opened 2020-12-17
Nans for pixels in reference window in linear_rate product


mcgarth:
**Observation**: Nans for pixels within reference window in linear_rate product

**Possible cause**:  are pixels with zero value set to Nan following the APS spatial filter operation?

**Solution**: during ref phase subtraction, assign ref phase to a sub-millimetric number (e.g. 0.00001) instead of zero.

chandra2ga:
This is an important issue and a bug in the program while processing pyrate with higher multi-look factor. 

**Concern**: When we used user selected reference pixel in case of higher multi-look factor e.g., 20 mlk. Some cases we encountered that after APS filtering step the program assign default NaN values to the reference pixel location (when not finding any neighboring ref pixel within a certain search window size) but then continue processing without showing any error. 

Here is an example of using Dsc Surat data: 20 mlk vs 15 mlk
If you look the zoomed view might not see much difference visually for both the velocity maps
![vel_map](https://user-images.githubusercontent.com/61173442/107281987-84374300-6aae-11eb-9cf8-51455cd1fecc.png)

But if we zoomed around the reference pixel location we will notice the actual differences and NaN pixel value assigned for 20 mlk velocity map. 
![zoomed_vel_map](https://user-images.githubusercontent.com/61173442/107282161-bc3e8600-6aae-11eb-97a6-2c3cd4ff2a7b.png)

**Solution**: Whenever program encountered such issue, program would automatically assign a non-zero small number of higher decimal (e.g., 0.000001) to the reference location.

richardt94:
I'm not sure if I agree with NaN-setting zero pixels after the APS if this is what is happening - surely we can create a mask from pixels that are NaN before the APS, do the APS (including interpolating the NaNs for compatibility with Fourier transform), and then use the mask to set no-data areas on the output? This seems more robust to avoid erroneously masking areas of the interferogram that happen to be close to zero.

mcgarth:
The immediate issue has been addressed with #335

PR #335, opened 2021-05-12
Fix issue 310

mcgarth:
This PR introduces a circumventive measure to address the Issue identified at: https://github.com/GeoscienceAustralia/PyRate/issues/310 

Specifically, I add a small value of 1e-20 to the reference phase that is subtracted from each interferogram. This avoids pixels taking the exact value of zero (0.0) and possibly being converted to NaN downstream in the workflow.


richardt94 (approved):
Looks good. You've added comments which make the reference phase correction process more clear in the code, as well as fixing the bug. Comments as noted - we'll eventually want a more robust fix in APS, up to you whether you want to flag this in comments as a "to do" for later. The offset strategy should work well enough for now.

basaks (approved):

Issue #323, opened 2021-03-22
Prepifg opens too many files


richardt94:
**Describe the bug**
The function `get_analysis_extent` in `prepifg_helper.py` iterates through a (sometimes very long) list of files and opens them all with GDAL through the `open` method of the `RasterBase` class. This sometimes results in conflict with the system's limit on the number of open files and results in a runtime error.

**To Reproduce**
1. Go to `/g/data/dg9/INSAR_ANALYSIS/SURAT_S1_Asc/S1/PYRATE/` on VDI.
2. Run `pyrate prepifg` using the config file in that folder.

**Expected behavior**
The `get_analysis_extent` functionality doesn't require all of the raster files to be open at the same time, so it should not be restricted by the system's limit on open file handles.

**Notes**
This can be resolved if we move file opening into the inner loop of `_get_same_bounds` (currently a list comprehension which is getting the geotransform of each object), and closing the dataset after we are done. I'm not sure if there are other instances in the PyRate workflow where all of the interferograms are opened at the same time, but if so these should be corrected in a similar fashion, and we should move to a more 'on-demand' file handling model to avoid this type of issue in the future.

richardt94:
Resolved by #336

PR #336, opened 2021-05-17
limit file open during prepifg

basaks:
To address https://github.com/GeoscienceAustralia/PyRate/issues/323

### DIFF
--- a/pyrate/core/prepifg_helper.py
+++ b/pyrate/core/prepifg_helper.py
@@ -76,10 +76,6 @@ def get_analysis_extent(crop_opt: int, rasters: List[Union[Ifg, DEM]], xlooks: i
             if not all([_is_number(z) for z in user_exts]):
                 raise PreprocessError('Custom extents must be 4 numbers')
 
-    for raster in rasters:
-        if not raster.is_open:
-            raster.open()
-
     _check_looks(xlooks, ylooks)
     _check_resolution(rasters)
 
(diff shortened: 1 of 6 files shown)


mcgarth (changes requested):
`get_analysis_extent` is also used in `pyrate/core/orbital.py` and `pyrate/core/refpixel.py`; these too will need to be modified

richardt94:
> 
> 
> `get_analysis_extent` is also used in `pyrate/core/orbital.py` and `pyrate/core/refpixel.py`; these too will need to be modified

Is this true? Unless steps in `orbital` and `refpixel` are relying on side effects of `get_analysis_extent` by assuming the ifgs are already open then those calls should still work OK

basaks:
yes, that is fine. Now on that function call we open the rasters as required and once they are used, the rasters are closed.
Now all the file open and close opeations are internal to get_analysis_extent childs, and the files are only open one at a time and closed asap.
Minimal files are open while using this function now.

richardt94 (approved):
Finally got VDI working to test the fix. Can confirm that the original issue with `prepifg` is fixed - no more complaints about file handles. I also see in the code that you've taken other steps to improve file handling.

mcgarth:
> > `get_analysis_extent` is also used in `pyrate/core/orbital.py` and `pyrate/core/refpixel.py`; these too will need to be modified
> 
> Is this true? Unless steps in `orbital` and `refpixel` are relying on side effects of `get_analysis_extent` by assuming the ifgs are already open then those calls should still work OK

Ah yes, no need to change the calls to `get_analysis_extent` in `pyrate/core/orbital.py` and `pyrate/core/refpixel.py`

mcgarth (approved):

PR #337, opened 2021-05-26
Raise exception when too many header files are provided

adeane-ga:
Very small PR to simply generate a clearer error message for when there are too many parameter files given to PyRate ini the lists. 

- `manage_headers()` in `pyrate/core/gamma.py` is designed to read and combine header information from a list of 2 MLI.par files or a single DEM.par file.
- The iff condition statement at `line 457` controls the flow: https://github.com/GeoscienceAustralia/PyRate/blob/c3054c569b4b1827b61326741277606bc47e8463/pyrate/core/gamma.py#L444-L471
- if the user accidently has more than 2 MLI.par files in their list (e.g. because of multiple polarisations), then this conditional statement will will not apply. 
- I have just added am `elif` condition to raise an error if the list of header files is > 2.

### DIFF
--- a/pyrate/core/gamma.py
+++ b/pyrate/core/gamma.py
@@ -463,6 +463,8 @@ def manage_headers(dem_header_file, header_paths, baseline_paths=None):
         else:
 #            baseline_header = {}
              combined_header = combine_headers(hdrs[0], hdrs[1], dem_header)
+    elif len(header_paths) > 2:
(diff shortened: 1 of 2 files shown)


adeane-ga:
Just added unit test to check if manage_headers() will throw exception when input IFG list has more than two files:

https://github.com/GeoscienceAustralia/PyRate/blob/75bc55be3ad885ad7296189177028943b9283dfc/tests/test_gamma.py#L178-L187

PR #338, opened 2021-06-01
[phase closure] sum closure memory opt [ci skip]

basaks:
memory optimisations during sum closure

### DIFF
--- a/pyrate/core/phase_closure/sum_closure.py
+++ b/pyrate/core/phase_closure/sum_closure.py
@@ -14,7 +14,7 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
-
+import resource
 from collections import namedtuple
 from typing import List, Dict, Tuple, Any
 from nptyping import NDArray, Float32, UInt16
@@ -24,6 +24,7 @@ import pyrate.constants as C
 from pyrate.core import mpiops
 from pyrate.core.shared import Ifg, join_dicts
 from pyrate.core.phase_closure.mst_closure import Edge, WeightedLoop
+from pyrate.core.logger import pyratelogger as log
 
 IndexedIfg = namedtuple('IndexedIfg', ['index', 'IfgPhase'])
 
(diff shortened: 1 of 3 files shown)


pyrate/core/phase_closure/sum_closure.py
```diff
@@ -88,23 +91,30 @@ def sum_phase_closures(ifg_files: List[str], loops: List[WeightedLoop], params:
...
         closure_dict = join_dicts(mpiops.comm.gather(closure_dict, root=0))
-        ifgs_breach_count_process = np.sum(np.stack(ifgs_breach_count_arr), axis=0)
-        ifgs_breach_count = mpiops.comm.reduce(ifgs_breach_count_process, op=mpiops.sum0_op, root=0)
+
+        total_gb = mpiops.comm.allreduce(ifgs_breach_count.nbytes / 1e9, op=mpiops.MPI.SUM)
+        log.info("Memory usage due to ifgs_breach_count {:2.4f}GB of data".format(total_gb))
+        log.debug(f"shape of ifgs_breach_count is {ifgs_breach_count.shape}")
+        log.debug(f"dtype of ifgs_breach_count is {ifgs_breach_count.dtype}")
+       
+        mpiops.comm.Reduce(ifgs_breach_count, ifgs_breach_count, op=MPI.SUM, root=0)
```
richardt94, line 114: I pushed a commit which changes this to a capital R reduce. This fixes further issues that lower-case reduce was having with creating a 13 GB pickle object to communicate its part of the `ifgs_breach_count_arr` back to the root node. Using `Reduce` is safe because while `mpiops.array_split` may result in different numbers of loops being sent to each process, the accumulated breach count arrays are all numpy arrays of the same size `(height of ifg, width of ifg, # of ifgs)` so we can use preallocated buffers for our reduction operation (and the sum over axis 0 is default behaviour of `MPI.SUM` so we don't need to mess with a custom `sum0_op`). Still a couple of things to change before we can merge this:
1. I'm not sure if it's safe to use the same buffer for sending and receiving for the reduction - if not we'll need to create an empty array of zeros
2. This commit will break the non-MPI option, because I had to directly import mpi4py in `sum_closure.py` to get `MPI.SUM`. Should be easy to fix this by exporting MPI.SUM out of mpiops.

basaks: good pick up on the `Reduce` use here Richard. Thank you.

richardt94:
Thanks Sudipta for your work debugging this PR and fixing up the buffer-based Reduce method. As far as I can tell it is now fully functional, so I'm happy to merge it now - might be worth getting @mcgarth 's opinion first

mcgarth (approved):

PR #339, opened 2021-06-10
Replace dictionaries with buffers when merging loop closure calculations between processes [ci skip]

richardt94:
Was having another issue with `sum_phase_closures` which was related to the numpy-array-in-a-dictionary design pattern we're using for inter-process communication - this simply doesn't work with big stacks of interferograms because if the limit on maximum size of a pickle, so I replaced it with a preallocated buffer and a call to `Gatherv`, along with the associated calculations of buffer sizes and stuff for each process.

### DIFF
--- a/pyrate/core/phase_closure/sum_closure.py
+++ b/pyrate/core/phase_closure/sum_closure.py
@@ -107,6 +107,9 @@ def sum_phase_closures(ifg_files: List[str], loops: List[WeightedLoop], params: 
         log.info("Memory usage due to ifgs_breach_count_process {:2.4f}GB of data".format(total_gb))
         log.debug(f"shape of ifgs_breach_count_process is {ifgs_breach_count_process.shape}")
         log.debug(f"dtype of ifgs_breach_count_process is {ifgs_breach_count_process.dtype}")
+
+        total_gb = mpiops.comm.allreduce(closure_process.nbytes / 1e9, op=mpiops.MPI.SUM)
+        log.info("Memory usage due to closure_process {:2.4f}GB of data".format(total_gb))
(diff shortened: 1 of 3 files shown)


pyrate/core/phase_closure/sum_closure.py
```diff
@@ -91,19 +91,37 @@ def sum_phase_closures(ifg_files: List[str], loops: List[WeightedLoop], params:
...
+            for i in range(1,mpiops.size):
+                displs.append(displs[-1] + counts[i-1])
+            ifgres = np.prod(ifgs[0].phase_data.shape)
+            counts = [c * ifgres for c in counts]
+            displs = [d * ifgres for d in displs]
+            mpiops.comm.Gatherv([closure_process, len(process_loops), mpiops.MPI.FLOAT],
+                                [closure, counts, displs, mpiops.MPI.FLOAT], root = 0)
+            #TODO use Gatherv instead of allocating extra zeros and using Reduce
+        else:
+            closure = closure_process #if MPI isn't installed there's only one process so this is fine
```
mcgarth, line 124: Is this true? If MPI is not installed, the user may still be using `multiprocessing` parallelism. Does the logic here catch the `multiprocessing` split of processes within the `if mpiops.MPI_INSTALLED:` if block (even if MPI is in fact not installed)?

richardt94: Is this the same as `joblib` parallel processing? If so, this is handled by the outer branch (line 85) - the only reason the code will reach line 124 is if MPI isn't installed and the user has turned parallelism off.

basaks (approved):

PR #341, opened 2021-06-17
Fix reference phase bug

richardt94:
@mcgarth noticed a bug where areas outside the frame contained non-NaN values in the phase closure loop plots. I traced this back to the reference phase subtraction - this was done without masking the no-data values in the source dataset, so if the no-data val was zero, no-data pixels would become non-zero after subtracting the reference phase, and then be treated as valid pixels downstream. This PR converts no-data pixels to NaN before doing ref phase subtraction.


mcgarth (approved):
I approve this PR subject to the test suite all passing

PR #343, opened 2021-06-18
Change method to drop ifgs with unwrapping errors in phase closure iterations

richardt94:
This PR changes the method we use to drop interferograms from the next iteration of phase closure checking - currently, a misclosure in a pixel for *any* loop the ifg appears in counts towards the comparison with `avg_ifg_err_thr`. If we have a large stack with many loops, an unwrapping error in one ifg will result in misclosures for many other ifgs, so this method means we end up dropping a lot of good interferograms.

The PR uses a stricter criterion - how many pixels have misclosures for *all* loops the ifg appears in? In my experiments, this results in keeping a lot more good ifgs for large stacks, and as a bonus is consistent with the logic we use to mask unwrapping errors in the final output of the phase closure module.


basaks (approved):

mcgarth (approved):
This change makes sense. Good pick up!

PR #344, opened 2021-06-21
Bugfix: Detect and ignore NaNs in per-pixel dem error calculation

adeane-ga:
_This is a small PR for some additions to the `dem_error.py` file, we should create a new PR for some proper testing of the DEM error corrections._

**The per-pixel DEM error estimation loop would fail because the loop would contain NaNs from out of frame regions**
- added a for loop skip in dem error estimation for pixels with NaNs

**The baseline metadata in the IFG GeoTIFFs were not being updated because of the use of np.mean in the presence of NaNs**
- converted 2x np.mean() functions to np.nanmean()

_I also did some code tidying_


mcgarth (approved):
Good pick ups!

PR #345, opened 2021-07-05
Add plotting script for correction files

adeane-ga:
This script is designed to plot the correction files (NumPy), against the original interferogram and the corrected interferogram (GeoTIFFs) to aid user's interpretation of corrections and help debugging with PyRate. 

Here is an example where from left to right is original IFG (`interferogram_dir/`), correction applied (`orbit_error_dir/`), and resulting corrected IFG (`temp_mlooked_dir/`):

image

**Help Output**

image

**Example Usage:**

`python3 ~/PyRate/utils/plot_correction_files.py ./interferogram_dir ./orbit_error_dir ./temp_mlooked_dir ./plot_save_dir 1 10`

_the last two arguments will indicate that it will plot from the 1st to the 10th IFG (10 separate PNGs), this can be changed to any range within the maximum number of IFGs in user PyRate directory. Using `1 1 ` will plot the 1st only_

**Notes**

- This script is designed to be generic and as simple as possible, so it is easy for any given user to change it on the fly in there own local repository (rather than a script that deals with every possible combination of processing). 
- The script is also naïve, in that it relies on the user to decide how valid the comparison is, depending on user's own processing (which corrections were applied etc.). 
- I have only tested this script on the Orbit and APS corrections, it theoretically should work on the DEM error correction too but I have not tested it. 
- Currently, the files we save out as APS "correction" files, don't seem to contain the full correction, work will be done to investigate this. 
- Users can plot a single IFG or the full time series based on the last two arguments (but users need to have awareness of how many IFGs are in their directory).
- If plotting full time series, consider running as a QSUB job (if on NCI), otherwise plotting subsets will not take up too much memory.

### DIFF
--- a/utils/plot_correction_files.py
+++ b/utils/plot_correction_files.py
@@ -112,26 +112,25 @@ for i in range(first_ifg_num, last_ifg_num + 1):
     ifg_quant = np.nanquantile(ifg, [0.05, 0.95])
     climit = np.nanmax(np.abs(ifg_quant))
     
-    # subtract median
-    
+    # switch to subtract median or not
     if subtract_median == 1:
-        sub_med = 1
+        switch = 1
     else:
-        sub_med = 0
+        switch = 0
     
 
     fig, ax = plt.subplots(1,3, figsize=(6, 3))
      
     # IFG
-    s0 = ax[0].imshow(ifg-(np.nanmedian(ifg)*sub_med), cmap='bwr', clim=(-1*climit, climit))
+    s0 = ax[0].imshow(ifg-(np.nanmedian(ifg)*switch), cmap='bwr', clim=(-1*climit, climit))
     ax[0].set_axis_off()
 
     # CORRECTION FILE
-    s1 =ax[1].imshow(corr-(np.nanmedian(corr)*sub_med), cmap='bwr', clim=(-1*climit, climit))
+    s1 =ax[1].imshow(corr-(np.nanmedian(corr)*switch), cmap='bwr', clim=(-1*climit, climit))
     ax[1].set_axis_off()
 
     # IFG CORRECTED
-    s2 = ax[2].imshow(ifg_corr-(np.nanmedian(ifg_corr)*sub_med), cmap='bwr', clim=(-1*climit,climit))
+    s2 = ax[2].imshow(ifg_corr-(np.nanmedian(ifg_corr)*switch), cmap='bwr', clim=(-1*climit,climit))
     ax[2].set_axis_off()
 
     # Extra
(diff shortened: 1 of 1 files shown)


utils/plot_correction_files.py
```diff
@@ -0,0 +1,124 @@
...
+for file in glob.glob(f'{corr_dir}/*.npy'):
+    corr_list.append(file)
+
+tempml_list = []
+for file in glob.glob(f'{tempml_dir}/*ifg.tif'):
+    tempml_list.append(file)
+
+# Sort
+ifg_list.sort()
+corr_list.sort()
```
richardt94, line 64: Are the contents of ifg_list, corr_list and tempml_list guaranteed to be from an identical set of date pairs? (i.e. do the corrections ever fail and result in no corrections being saved for one or more of the ifgs?)

If not, the sort-all-lists approach here may result in mismatching ifgs to corrections in the plots.

utils/plot_correction_files.py
```diff
@@ -0,0 +1,124 @@
...
+    date_pair_string = date_pair_list[0]
+    
+    print(f'\nPlotting for {date_pair_string}...\n')
+    
+    # Plot
+    climit = 100
+    fig, ax = plt.subplots(1,3, figsize=(6, 3))
+     
+    # IFG
+    s0 = ax[0].imshow(ifg-np.nanmedian(ifg), cmap='bwr', clim=(-1*climit, climit))
```
richardt94, line 96: I get that we're subtracting the median so that the zero-centred colour bar will work, but it might be better not to do this, as if there's some constant bias on the correction or the input ifg this won't show up in the plots. You could instead adaptively figure out a colour range using quantiles of the inputs, then plot each figure on this same colour range.

richardt94 (approved):
Thanks Al, this is good functionality to have for debugging our outputs. I have a couple of comments on ways this could be improved - I'm approving the PR, so they're not mandatory, but please consider them before merging.

utils/plot_correction_files.py
adeane-ga: Yeah good point, I look into this before merging.

mcgarth (approved):
Thanks Al, great contribution and exceptionally documented!

Before merging, please just check that al the script's package dependencies are in the `requirements-plot.txt`

utils/plot_correction_files.py
adeane-ga: Cool, so I've 

- added in an optional argument to subtract the median from the plots or not (default is to subtract median)
- I've also changed to adaptive colour scale to use the maximum value out of the absolute value from the 5% and 95% quantile of the interferogram data.

utils/plot_correction_files.py
adeane-ga: This change has been made by comparing the date-pairs of the three datasets per figure.

PR #346, opened 2021-07-06
Change which file is saved to aps_error_dir

adeane-ga:
After using the correction plotting script on the APS error correction directory (`aps_error_dir`), it was apparent that the file saved out as a "correction", is actually the corrected IFG file. 

At line 97 in the `spatio_temporal_filter()` function below we see that `tsincr` is the time series epoch corrected for APS, and the `ts_aps` is the combined temporal and spatial filter which is subtracted from the data to perform the correction. At line 173 and 174 inside `_ts_to_ifgs()`, we can see that it is the corrected phase that is saved both to the `tmp_dir` and also the `aps_error_dir`. Here we should be saving `ts_aps` into the `aps_error_dir`. This PR is for making this change. 

https://github.com/GeoscienceAustralia/PyRate/blob/508fb4b2a0b5fb2f53ad2b3cf508c7b6db37956a/pyrate/core/aps.py#L79-L100

https://github.com/GeoscienceAustralia/PyRate/blob/508fb4b2a0b5fb2f53ad2b3cf508c7b6db37956a/pyrate/core/aps.py#L153-L174

To summarise, below are some plots output from a PyRate process with only the APS error correction applied. The only difference between plots is which file is being saved out to the `aps_error_dir`.

**Plot structure**
_Left:                 Uncorrected Interferogram
Middle:           Correction File
Right:              Corrected Interferogram_

### **What we are currently saving out to aps_error_dir:**

image

### **This PR will result in the appropriate correction file being saved in  aps_error_dir, as seen by this plot:**

image

### **To confirm, I made a 4th panel which is MIDDLE + RIGHT to see if we get back to the original interferogram:**

image

**An interesting note**
Out of interest, its also worth noting line 172 (below), where (I think) the corrected time series is being reconstructed into the original IFGs by datepair by summing the cumulative time series between two associated epochs of the corrected phase. A potential issue with this, is that whatever is not captured by the APS spatiotemporal filter will propagate through into the listed IFGs upon reconstruction. This may explain why we are getting "ghost" tropospheric features appear in unrelated IFGs. I think this is what Richard was saying when we first noted them. I may be wrong, but thought I'd note it for discussion.

https://github.com/GeoscienceAustralia/PyRate/blob/508fb4b2a0b5fb2f53ad2b3cf508c7b6db37956a/pyrate/core/aps.py#L172

### DIFF
--- a/pyrate/core/aps.py
+++ b/pyrate/core/aps.py
@@ -96,7 +96,7 @@ def spatio_temporal_filter(tsincr: np.ndarray, ifg_paths: List[str], params: dic
     ts_aps = spatial_low_pass_filter(ts_hp, ifg, params)
     tsincr -= ts_aps
 
-    _ts_to_ifgs(tsincr, preread_ifgs, params)
+    _ts_to_ifgs(tsincr, preread_ifgs, params, ts_aps)
     ifg.close()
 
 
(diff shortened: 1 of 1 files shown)


mcgarth (commented):
In reviewing this PR I have found that there a much bigger issues to be addressed, beyond the scope of just the APS module. I am therefore closing this PR and will include the single commit from here in https://github.com/GeoscienceAustralia/PyRate/pull/347

PR #366, opened 2021-10-20
GitHub Pages action

sixy6e:
Some minor changes to get the GitHub Pages action working.  The only page that didn't display any content (apart from a heading) is the `Ifconstats Module` so some refinement is probably still required.

The `conf.py` had to be modified as the commonmark module was causing conflicts with an existing markdown parser. So it could be marked for removal.


mcgarth:
Thanks for the improvements @sixy6e!

Issue #362, opened 2021-10-18
Error Handling: error raise message is incorrect


adeane-ga:
**Very simple problem and fix**
Line 315 of `prepifg_helper.py` has an error exception message that states the wrong case.

https://github.com/GeoscienceAustralia/PyRate/blob/8203989e7d9b675491e79610cff04728d8dc3973/pyrate/core/prepifg_helper.py#L314-L317

Should state:

`ifgxfirst must be less than ifgxlast`

adu461386118:
agree, ifgxfirst and ifgxlast are two longitude values, ifgxfirst should be smaller than ifgxlast,

adeane-ga:
Closing this issue because it was fixed with this merge into develop: https://github.com/GeoscienceAustralia/PyRate/pull/369

Thanks.

Issue #359, opened 2021-10-18
Metadata addition: LOS direction in GeoTIFF metadata


adeane-ga:
**Problem**
When users are interpreting PyRate velocity outputs, it would be useful to quickly be able to query in which direction surface motion can/cannot be observed based on the direction of the LOS observations. Currently a user has to query the `azimuth.tif` file to obtain this information. It would be convenient if one could access the median azimuth angle in the metadata of the GeoTIFFs within the `velocity_dir`. 

**Request**
- have the median azimuth angle of the LOS observation placed into the metadata of the GeoTIFF files in PyRate output. 
- In particular these files:

`velocity_dir/linear_rate.tif`
`velocity_dir/linear_error.tif`
`velocity_dir/linear_rsquared.tif`
`velocity_dir/linear_samples.tif`
`velocity_dir/linear_error.tif`

But potentially in the time series files within `timeseries_dir` directory too.

adeane-ga:
> I believe the average incidence angle is already saved to the interferogram geotiff metadata?

Oh, here I meant the average azimuth angle (azimuth direction of target to sensor). 

My reasoning for the additional metadata here is just so non-experts can have something to refer to independently when investigating the velocity file and have specific infrastructure in question where they may already know the expected direction of motion and whether it may or may not be observed in their velocity file due to the sensing geometry.

adeane-ga:
Closing this issue because it was fixed with this merge into develop: https://github.com/GeoscienceAustralia/PyRate/pull/369

Issue #358, opened 2021-10-18
Metadata addition: reference pixel coordinates to metadata in output files in velocity directory


adeane-ga:
**Problem**
Outside users of PyRate products often only engage with or have access to the results from the velocity directory `velocity_dir`. The files in there are:
- `linear_error.tif`
- `linear_rate.tif`
- `linear_rsquared.tif`
- `linear_samples.tif`

None of these files have the reference pixel coordinates in the GeoTIFF metadata. 

For any practical interpretation of PyRate results, a user needs to know where the reference pixel is. 

Currently it is within the metadata of the time series files (`tscuml_YYYYMMDD-YYYYMMDD.tif` files in `timeseries_dir`), but practically users may only have access to the ultimate velocity results and not engage with time series data. 

**Request**
Simply place the reference pixel coordinates into the GeoTIFF metadata of the files that are output into the velocity directory (like it has been for the time series GeoTIFFs).

adeane-ga:
> I believe the reference pixel coordinates are already saved to the interferogram geotiff metadata, in both X/Y and lat/lon notation.

This is a strange one, because reference pixel metadata is not currently being saved to the GeoTIFFs in the velocity directory. But in some past solutions it has been. So I am not sure what's changed, and I cannot find the section of code that does add reference pixel metadata to the velocity directory files - this current PR is working on it https://github.com/GeoscienceAustralia/PyRate/pull/369.

But, yes the reference pixel metadata is currently being saved to the interferogram and time series GeoTIFFs. This addition is for users who are only looking at the velocity directory files.

adeane-ga:
Closing issue because it was fixed with this merge into develop: https://github.com/GeoscienceAustralia/PyRate/pull/369

PR #369, opened 2021-10-22
Add metadata to output files

adeane-ga:
This is a small PR that adds some simple metadata to the output files in the velocity directory. 

**Specifically**
Here are the three changes and their associated GitHub issues:

1. Add latitude and longitude coordinates of reference pixel to the output GeoTIFFs in `velocity_dir` (https://github.com/GeoscienceAustralia/PyRate/issues/358).

2. Add median azimuth of the LOS vector to the output GeoTIFFS in `velocity_dir` (https://github.com/GeoscienceAustralia/PyRate/issues/359).

3. Just a small correction to an error message which was stating the wrong condition (https://github.com/GeoscienceAustralia/PyRate/issues/362)

In addition to checking that none of these break your processing workflow, here are some things to check regarding  the above points. 

1. Confirm that the reference pixel coordinates are in the output GeoTIFF metadata at the end of processing.
2. At the end of processing, check that the median azimuth is in the output GeoTIFF metadata, or that `NOT COMPUTED` is there in the case that the geometry files are not created. 
3. Just read over the code and confirm if the conditional logic is correct.

### DIFF
--- a/pyrate/core/prepifg_helper.py
+++ b/pyrate/core/prepifg_helper.py
@@ -313,7 +313,7 @@ def _custom_bounds(ifgs, xw, ytop, xe, ybot):
 
     if xe < xw:
         raise PreprocessError('ERROR Custom crop bounds: '
-                              'ifgxfirst must be greater than ifgxlast')
+                              'ifgxfirst must be less than ifgxlast')
 
     for par, crop, orig, step in zip(['x_first', 'x_last', 'y_first', 'y_last'],
                                      [xw, xe, ytop, ybot],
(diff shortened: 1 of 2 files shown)


adeane-ga:
GeoTIFF metadata that is viewable with `gdalinfo` in Linux, is still not viewable with QGIS GUI interface `right click on raster layer => properties => information`. 

Here is one thread which looks like a similar problem, but I still can't determine the issue from it:
https://gis.stackexchange.com/questions/214247/set-raster-metadata-in-gdal-and-view-in-gis-software

This is yet to be resolved, so I guess we could keep this PR open until we figure it is resolved, or we could merge and then list the metadata discrepancy in the issues page so we can come back to it. 

Not sure of the best action.

adeane-ga:
One thought:

is there a difference between using `SetMetadata()` compared to `gdal_edit()`?

adu461386118:
I confirm I can see the reference pixel coordinates within the metadata of output files in velocity directory, also, I think both SetMetadata() and gdal_edit() can do pretty much the same thing, that is to set/edit various information of an existing GDAL dataset, don't really know the difference between them,

s-m-t-c:
I'm going to merge this and document the qgis issue.

PR #376, opened 2022-01-24
Added list creator

s-m-t-c:
This is a small helper script that makes input lists for PyRate based on the output of Gamma.
It also prints the number of files in each category (interferograms, coherence files etc).

If you have a similar script that offers improvements feel free to modify this PR.

### DIFF
--- a/pyrate/core/prepifg_helper.py
+++ b/pyrate/core/prepifg_helper.py
@@ -313,7 +313,7 @@ def _custom_bounds(ifgs, xw, ytop, xe, ybot):
 
     if xe < xw:
         raise PreprocessError('ERROR Custom crop bounds: '
-                              'ifgxfirst must be less than ifgxlast')
+                              'ifgxfirst must be greater than ifgxlast')
 
     for par, crop, orig, step in zip(['x_first', 'x_last', 'y_first', 'y_last'],
                                      [xw, xe, ytop, ybot],
--- a/pyrate/core/refpixel.py
+++ b/pyrate/core/refpixel.py
@@ -20,7 +20,6 @@ of the interferometric reference pixel
 import os
 from os.path import join
 from typing import Tuple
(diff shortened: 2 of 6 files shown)


adeane-ga:
- I've had a look at the PR and can confirm that this functionality does not exist yet, and so it provides a useful addition for users.
- I haven't tested the the script, but it does magic logical sense and shouldn't cause any issues.

mcgarth:
Hi @s-m-t-c, was there a reason you needed to pin a version to GDAL? I remember that we left this blank because the version gets found and updated based on system GDAL found during running of `python setup.py install`. @basaks might remember more...

s-m-t-c:
@mcgarth I was actually just trying to get the CI checks to pass and thought it might be because the GDAL version had changed. This doesn't seem to solve the problem though so I will revert it. Any ideas on why they're failing?

mcgarth:
I think the CI fails are not related to the script addition you have made here. The script is not covered by the test suite (nothing in `utils` dir is covered, and doesn't need to be IMO). I think you can safely revert your branch here back to [3a31a6b](https://github.com/GeoscienceAustralia/PyRate/pull/376/commits/3a31a6bb4fe6e72027f8ee903695d94d88632df5) and merge to `develop` despite the CI fail. After the merge, if `develop` is failing then there will need to be a separate investigation as to why (seems to be related to gdal not getting imported in to test envirnoment)

adu461386118:
-For multi-frames: We currently have a function - pyrate_prep.py. this discovers interferograms and auxiliary files in dz56 for multi-frames and produces pyrate config files and lists in a specified directory. The location is in: /g/data/dg9/INSAR_ANALYSIS/NSW_mosaic_final_2/pyrate_prep.py, 
-For individual frame: I have tested the the script, and confirm it is working properly by providing path to the Gamma folder titled with the frame name

PR #382, opened 2022-03-15
Only run CI on master and develop

richardt94:
The current `build.yml` runs on pushes to all branches, including the `gh-pages` branch which only contains the docs and no actual code, and will therefore always fail CI. It makes more sense to run it only on the "main" branches (`master` and `develop`), especially given the resources consumed by the CI runs.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,9 +5,11 @@ on:
   pull_request:
     branches:
       - 'master'
+      - 'develop'
   push:
     branches:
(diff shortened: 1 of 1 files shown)


richardt94:
I also updated the CI workflow names so they're not just all called "Python" - this will make debugging easier if things break on specific python/gdal versions.

mcgarth (approved):
Changes make sense. @adu461386118 to confirm and merge

